PR #145, opened 2014-06-30
Bug 1031551 - Tasks go below search filter unless page is maximized

bobsilverberg:
This is awaiting a review from @rbillings


bobsilverberg:
@rbillings I have finally gotten around to getting this ready for you to review again. You can see this version at https://oneanddone-dev-pr-145.herokuapp.com/en-US/. It may become irrelevant with the new design (which likely won't happen until 2016 anyway), but I think it's still worth implementing for now.

rbillings:
This change looks good. It may become irrelevant next year, but it makes the current site a lot nicer.

bobsilverberg:
Thanks @rbillings. Due to the Heroku pipeline, you should see this change automatically show up on dev within a few minutes. You could verify it there and let me know if it's good, and then I can push to stage and prod.

PR #166, opened 2014-07-27
Bug 1032759 - [testing] Test Filter Tasks According to Estimated Time

bitgeeky:
Added required css selector.


bobsilverberg:
As we are no longer looking for this test (plus the filter is going to change too), I am going to close this PR.

PR #190, opened 2014-10-02
Test ImportTasksView

mjzffr:
Any feedback before I continue, @bobsilverberg?

My comment about these from previous PR: 
"I added some tests so y'all can see where I'm headed and maybe steer me in the right direction in terms of using Mock and such. I was in the middle of writing tests to see if the right template is used depending on which "stage" is posted. I think I need to pass in complete form data to get those tests to work."


bobsilverberg:
Thanks @mjzffr. I finally got around to reviewing this again. I had to make a few changes due to the upgrade to Django 1.7, but I maintained your ownership of the commit. These changes landed in https://github.com/mozilla/oneanddone/commit/ca9b749fda78e7af728dd88d0137593bcf01ff08

mjzffr:
Whoa, cool. :) Thanks @bobsilverberg :sunflower:

PR #227, opened 2014-12-19
Fix Bug 1096353 - Personal url should open in a new tab/window

tessie:


bobsilverberg:
Looks good. Thanks @tessie and @glogiotatidis!

PR #228, opened 2014-12-19
Bug 1096351 - User should view profile after editing 

tessie:


oneanddone/users/views.py
```diff
@@ -116,12 +116,12 @@ def get_object(self, *args, **kwargs):
 class UpdateProfileView(LoginRequiredMixin, generic.UpdateView):
     model = UserProfile
     template_name = 'users/profile/edit.html'
-    success_url = reverse_lazy('base.home')
+    success_url = reverse_lazy('users.profile.mydetails')
```
bobsilverberg: This is good @tessie! Thanks for taking on this task.

One thing I noticed when reviewing the code is that we don't actually need to specify the `success_url` here as we are doing a specific redirect to  `users.profile.mydetails` in the `form_valid` method. That is overriding the `form_valid` method in the generic view which is the only place where `success_url` is used. I realize that this code was there before you made your change, but if we see areas that can be improved when we are working on a module we should take the opportunity to do that.

Therefore, please remove this line of code, and if everything still works (it did when I tested it), then we will merge the changes with that code removed.

tessie:
@bobsilverberg .Fixed :)

bobsilverberg:
Great work! Thanks @tessie and @glogiotatidis. Merging this!

PR #230, opened 2014-12-22
 	Bug 1086626 - Task owner dropdown should sort alphabetically

tessie:


bobsilverberg:
You made the changes in the correct places, @tessie, but the sort field isn't correct. What displays in the dropdown is the result of the `__unicode__` method [1] of the `User` object, which in turn calls the `display_name` property [2] of the `User` object, which uses the `name` property of the `profile` [3]. So what we really want to sort on is the `name` of the associated `UserProfile`, which you can get by specifying `profile__name` as your sort field.  For example:

`queryset=User.objects.filter(task__isnull=False).distinct().order_by('profile__name'))`

The way it is coded right now, it will just the the `User.username` property which is not what is displayed, so that leads to the dropdown not being sorted alphabetically. I could see this on my system when I ran the code with your changes, but maybe you have fewer users on your system so didn't see this result.

Please update the sort field and this PR should be good to go.

[1] https://github.com/mozilla/oneanddone/blob/master/oneanddone/users/models.py#L24
[2] https://github.com/mozilla/oneanddone/blob/master/oneanddone/users/models.py#L45
[3]https://github.com/mozilla/oneanddone/blob/master/oneanddone/users/models.py#L51

tessie:
@bobsilverberg .will fix it soon

tessie:
@bobsilverberg .Have made the changes.sorted using profile name

bobsilverberg:
Looks good, @tessie. Please squash the commits and then I will merge. Thanks!

tessie:
@bobsilverberg .squashed

bobsilverberg:
Great! Thanks @tessie. :)

PR #231, opened 2014-12-22
 Fix Bug 1096431 - Able to create tasks with duplicate names

tessie:


bobsilverberg:
I see this was replaced with PR #239. I will close this. In the future, @tessie, if you create a new PR to replace an existing one you can close the old one yourself. Even better would be to not create a new PR, and just replace the code in the old one. This makes it easier to see all of the comments that have been added along the way.

PR #233, opened 2014-12-23
Fixed Bug 1032426 - Unclear that front page displays only some tasks

tessie:
@glogiotatidis  or @akatsoulas  Please review my PR.https://bugzilla.mozilla.org/show_bug.cgi?id=1032426

Thank you in advance

### DIFF
--- a/oneanddone/base/views.py
+++ b/oneanddone/base/views.py
@@ -4,6 +4,8 @@
 from django.shortcuts import redirect
 from django_filters.views import FilterView
 
+from tower import ugettext as _
+
 from oneanddone.tasks.filters import TasksFilterSet
 from oneanddone.tasks.models import Task
 from oneanddone.tasks.mixins import TaskMustBeAvailableMixin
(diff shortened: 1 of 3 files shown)


bobsilverberg:
Thanks for doing this @tessie. It isn't quite perfect though. The bug only wanted us to change the heading for the home page, but this change makes the heading change on the home page [1] and also on the available tasks page [2] and the last listing page [3].

Can you please update it so the heading only changes on the home page [1]?

[1] http://127.0.0.1:8000/en-US/
[2] http://127.0.0.1:8000/en-US/tasks/available/
[3] http://127.0.0.1:8000/en-US/tasks/list/

tessie:
@bobsilverberg .Please review my PR.I have bassed a the title as a context variable so that heading changes only on the home page

bobsilverberg:
This is good @tessie, but I would suggest that just using the name `header` for the variable could be a bit confusing as a header could mean a number of things, and is often used to refer to a whole area of a page. I suggest changing the name of the variable to `task_list_heading`, which is much more specific.

tessie:
@bobsilverberg .squashed

tessie:
thanks for reviews

bobsilverberg:
Excellent! Thanks @tessie!

PR #234, opened 2014-12-23
Bug 1050418 - Remove ability to list and get details of Users via the One .e and Done User REST API

tessie:


bobsilverberg:
Thanks for taking this on @tessie. There is a problem with this PR in that the changes are causing a number of tests to fail (as Travis is telling us above). Please run the unit tests locally before submitting a PR and make sure all tests are passing. Also, if you submit a PR and see that Travis is failing, then you should investigate that and try to fix those problems.

I'm not sure that the approach you took in this PR is entirely correct, but the first step is to resolve any test failures anyway, so I will leave that to you. Once there are no failures I will review the PR for correctness.

tessie:
actually I forgot to close the PR.I am sorry

bobsilverberg:
@tessie Okay, so are you going to close this now?

PR #235, opened 2014-12-26
Bug 1096431 - Able to create tasks with duplicate names

skjindal93:
Added "unique = True" in the models.py while creating a task for unique names.


bobsilverberg:
Thanks for taking this on, @skjindal93, but it looks like @tessie had already started working on this before you. See pull request #231 and later pull request #239.

This is a great effort, and we really appreciate the time you put into it, but it looks like a fix was already in the works. It might be a good idea to look at what pull requests are currently open when looking for a task to work on, so you don't end up doing duplicate work in the future.

PR #236, opened 2014-12-28
Bug 1036928 - Allow to set longer time estimages as 60 minutes

skjindal93:
I have added a new field called admin_time. I am just having problems of showing the value stored in database in the textfield. As in setting the initial value of admin_time if the value is already present in database.

### DIFF
--- a/oneanddone/tasks/forms.py
+++ b/oneanddone/tasks/forms.py
@@ -46,6 +46,13 @@ class PreviewConfirmationForm(forms.Form):
 
 
 class TaskForm(forms.ModelForm):
+    admin_time = forms.CharField(
+                help_text=_('Enter If more than 60 minutes.'),
+                required=False,
+                widget=forms.TextInput(attrs={'class': 'fill-width'}))
+    execution_time = forms.IntegerField(
+		widget=forms.Select(choices=[(0,'---')]+[(i, i) for i in (15, 30, 45, 60)]),
+		required=False)
     keywords = (forms.CharField(
                 help_text=_('Please use commas to separate your keywords.'),
                 required=False,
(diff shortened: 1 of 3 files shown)


bobsilverberg:
Thanks for taking this on @skjindal93. Great work thus far! I think you have made some really good progress here, but it's not exactly the way we'd like it to work. I would suggest taking a look at https://github.com/mozilla/oneanddone/pull/211 as someone else spent a lot of time on this and I answered many of their questions. This will give you a good idea of what we are looking for.

To summarize what's there (although it will still be useful for you to read through the code and comments on that PR):
- I don't think we need a new model field. We should be able to store the time in the existing `execution_time` field as it is an integer field.
- We will need a form field to collect this data from the user though.
- On the form, I think it would be best to keep the current `Estimated time` dropdown, but to add an option (perhaps at the bottom) for `> 60`. If a user chooses that option from the dropdown, then make an additional field appear which will allow them to enter the value.
- In terms of displaying the form for an existing Task, you'd need the same logic: display the extra field if the `execution_time` is already set to something greater than 60.

Does that all make sense?

rbillings:
@skjindal93 are you willing to finish this up with the comments above? If we don't hear from you in a week or so we will close the PR and update the bug assigned to you.

skjindal93:
@rbillings I got busy with my exams and assignment deadlines. I tried to resume to the work today but i am totally unable to run the server. I am getting weird errors. I made a fresh install of dJango a few weeks back for some of my work. I don't know may be the fresh install of dJnago is causing problems or not! I'll try to run the oneanddone server and have a look at the bug.I'll update if I am unable to run the server.

bobsilverberg:
@skjindal93 Do you expect to be able to work on this again soon? If not we will close this PR.

skjindal93:
No! I would not be able to work on this.

rbillings:
Closing

PR #237, opened 2014-12-29
Fix Bug 1065442 - Audit views that use PrivacyPolicyRequiredMixin as it  may no longer be required in some situations

tessie:
...may no longer be required in some situations

### DIFF
--- a/oneanddone/users/views.py
+++ b/oneanddone/users/views.py
@@ -157,13 +157,11 @@ class Verify(django_browserid.views.Verify):
         return super(Verify, self).login_failure(*args, **kwargs)
     
     def login_success(self, *args, **kwargs):
-	super(Verify, self).login_success(*args, **kwargs)
         if (not UserProfile.objects.filter(user=self.request.user).exists()):
-            return redirect('users.profile.create')
-        elif (not UserProfile.objects.get(user=self.request.user).privacy_policy_accepted):
-            return redirect('users.profile.update')
-        else:
-            return super(Verify, self).login_success(*args, **kwargs)
+            success_url = reverse_lazy('users.profile.create')
+        if (not UserProfile.objects.get(user=self.request.user).privacy_policy_accepted):
+            success_url = reverse_lazy('users.profile.update')
+        return super(Verify, self).login_success(*args, **kwargs,success_url)
 
 
 class UserDetailAPI(generics.RetrieveUpdateDestroyAPIView):
--- a/oneanddone/users/views.py
+++ b/oneanddone/users/views.py
@@ -155,6 +155,16 @@ class Verify(django_browserid.views.Verify):
             logging in, let us know by emailing <a href="mailto:{email}">{email}</a>.
         """).format(email='oneanddone@mozilla.com'), extra_tags='safe')
         return super(Verify, self).login_failure(*args, **kwargs)
+    
+    def login_success(self, *args, **kwargs):
+        if (self.request.user.is_authenticated() and
+                not UserProfile.objects.filter(user=self.request.user).exists()):
+            return redirect('users.profile.create')
+        elif (self.request.user.is_authenticated() and
+                not UserProfile.objects.get(user=self.request.user).privacy_policy_accepted):
+            return redirect('users.profile.update')
+        else:
+            return super(Verify, self).login_success(*args, **kwargs)
 
 
 class UserDetailAPI(generics.RetrieveUpdateDestroyAPIView):
--- a/oneanddone/users/views.py
+++ b/oneanddone/users/views.py
@@ -157,10 +157,10 @@ class Verify(django_browserid.views.Verify):
         return super(Verify, self).login_failure(*args, **kwargs)
     
     def login_success(self, *args, **kwargs):
-        super(Verify, self).login_success(*args, **kwargs)
-        if (UserProfile.objects.filter(user=self.request.user).exists()):
+	super(Verify, self).login_success(*args, **kwargs)
+        if (not UserProfile.objects.filter(user=self.request.user).exists()):
             return redirect('users.profile.create')
-        elif (UserProfile.objects.get(user=self.request.user).privacy_policy_accepted):
+        elif (not UserProfile.objects.get(user=self.request.user).privacy_policy_accepted):
             return redirect('users.profile.update')
         else:
             return super(Verify, self).login_success(*args, **kwargs)
--- a/oneanddone/users/views.py
+++ b/oneanddone/users/views.py
@@ -156,14 +156,14 @@ class Verify(django_browserid.views.Verify):
         """).format(email='oneanddone@mozilla.com'), extra_tags='safe')
         return super(Verify, self).login_failure(*args, **kwargs)
     
-    def login_success(self, *args, **kwargs):
-        if (not UserProfile.objects.filter(user=self.request.user).exists()):
-            self.success_url = reverse_lazy('users.profile.create')
-            return super(Verify, self).login_success(*args, **kwargs)
-        elif (not UserProfile.objects.get(user=self.request.user).privacy_policy_accepted):
-            self.success_url = reverse_lazy('users.profile.update')
-        self.success_url = reverse_lazy('base.home')
-        return super(Verify, self).login_success(*args, **kwargs)
+    @property
+    def success_url(self):
+        if not UserProfile.objects.filter(user=self.user).exists():
+            return reverse_lazy('users.profile.create')
+        elif not UserProfile.objects.get(user=self.user).privacy_policy_accepted:
+            return reverse_lazy('users.profile.update')
+        else:
+            return super(Verify, self).success_url
 
 
 class UserDetailAPI(generics.RetrieveUpdateDestroyAPIView):
--- a/oneanddone/users/views.py
+++ b/oneanddone/users/views.py
@@ -157,11 +157,10 @@ class Verify(django_browserid.views.Verify):
         return super(Verify, self).login_failure(*args, **kwargs)
     
     def login_success(self, *args, **kwargs):
-        if (self.request.user.is_authenticated() and
-                not UserProfile.objects.filter(user=self.request.user).exists()):
+        super(Verify, self).login_success(*args, **kwargs)
+        if (UserProfile.objects.filter(user=self.request.user).exists()):
             return redirect('users.profile.create')
-        elif (self.request.user.is_authenticated() and
-                not UserProfile.objects.get(user=self.request.user).privacy_policy_accepted):
+        elif (UserProfile.objects.get(user=self.request.user).privacy_policy_accepted):
             return redirect('users.profile.update')
         else:
             return super(Verify, self).login_success(*args, **kwargs)
--- a/oneanddone/users/views.py
+++ b/oneanddone/users/views.py
@@ -158,10 +158,12 @@ class Verify(django_browserid.views.Verify):
     
     def login_success(self, *args, **kwargs):
         if (not UserProfile.objects.filter(user=self.request.user).exists()):
-            success_url = reverse_lazy('users.profile.create')
-        if (not UserProfile.objects.get(user=self.request.user).privacy_policy_accepted):
-            success_url = reverse_lazy('users.profile.update')
-        return super(Verify, self).login_success(*args, **kwargs,success_url)
+            self.success_url = reverse_lazy('users.profile.create')
+            return super(Verify, self).login_success(*args, **kwargs)
+        elif (not UserProfile.objects.get(user=self.request.user).privacy_policy_accepted):
+            self.success_url = reverse_lazy('users.profile.update')
+        self.success_url = reverse_lazy('base.home')
+        return super(Verify, self).login_success(*args, **kwargs)
 
 
 class UserDetailAPI(generics.RetrieveUpdateDestroyAPIView):
--- a/oneanddone/base/views.py
+++ b/oneanddone/base/views.py
@@ -14,7 +14,7 @@ from oneanddone.users.models import UserProfile
 
 class HomeView(TaskMustBeAvailableMixin, FilterView):
     template_name = 'base/home.html'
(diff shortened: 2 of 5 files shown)


tessie:
@akatsoulas .Thanks squashed

bobsilverberg:
Thanks for doing this, @tessie. This does remove the mixin successfully, but I am a bit worried about guaranteeing that a user has signed the privacy policy. The way this is currently guaranteed is by the fact that, after a user logs in, they are directed to the Home page.

The view that serves the Home page has a `dispatch` method [1] which forces the user to create a profile and/or sign the privacy policy before proceeding.

The problem is that users are not always redirected to the home page, so this check will not always be performed.  For example, if a user clicks the Get Started button on a task without being logged in, and then logs in, they are directed directly to that task.

So we actually need to have the check that a user has signed the privacy policy to always be made after logging in, not just at the Home page. I believe the way to do this would be to extend the `login_success` method of django_browserid [2], which we would do by adding our own `login_success` method to our `users.Verify` view [3].

@tessie Is this something you want to try to do?

[1] https://github.com/mozilla/oneanddone/blob/master/oneanddone/base/views.py#L20
[2] https://github.com/mozilla/oneanddone/blob/master/vendor-local/lib/python/django_browserid/views.py#L50
[3] https://github.com/mozilla/oneanddone/blob/master/oneanddone/users/views.py#L151

tessie:
@bobsilverberg .I would like to give a try

tessie:
@bobsilverberg .I am attempting to fix it.

bobsilverberg:
@tessie This is a good start, but it's not actually working because there are a number of problems with it. Did you test it locally on your machine? If you tested the scenario where one would expect the user to be redirected to the Profile page you would have seen, as I have, that it's not working. It's important to make sure that you test your code before submitting a pull request as you shouldn't submit code unless you know it works.

I will add some comments to the code about the problems.

oneanddone/users/views.py
```diff
@@ -151,6 +151,16 @@ def login_failure(self, *args, **kwargs):
             logging in, let us know by emailing <a href="mailto:{email}">{email}</a>.
         """).format(email='oneanddone@mozilla.com'), extra_tags='safe')
         return super(Verify, self).login_failure(*args, **kwargs)
+    
+    def login_success(self, *args, **kwargs):
+        if (self.request.user.is_authenticated() and
```
bobsilverberg: One of the reasons this is not working is that `self.request.user.is_authenticated()` will always return `False` here because the user has not been authenticated yet. You are calling this code before the parent `Verify` has had a chance to authenticate the user. Therefore, you need to remove this first part of the `if` statement from here and also from line 159. 

Because this is `login_success` we can assume that the user is about to be authenticated, so we don't need to check for that at all.

oneanddone/users/views.py
```diff
@@ -151,6 +151,16 @@ def login_failure(self, *args, **kwargs):
...
+                not UserProfile.objects.filter(user=self.request.user).exists()):
+            return redirect('users.profile.create')
+        elif (self.request.user.is_authenticated() and
+                not UserProfile.objects.get(user=self.request.user).privacy_policy_accepted):
+            return redirect('users.profile.update')
+        else:
+            return super(Verify, self).login_success(*args, **kwargs)
```
bobsilverberg: If you only call `super(Verify)` here, then in the first two cases, where you want to redirect the user, they will not end up being authenticated at all, because that happens in the parent `Verify`. Also, as you will see if you remove the `if` statements I mentioned above, not calling the parent `Verify` will result in an error.

You need to call the parent `Verify` in all cases, and that method is what will do the redirect for you. Take a look at that to see what you need to do in this method before calling `super(Verify)` in order to redirect to the page in question.

bobsilverberg:
In addition to my comments above, once you get this working properly with `login_success`, we can remove the code in the home page that does the redirection as it will always be done on login.

oneanddone/users/views.py
```diff
@@ -151,6 +151,15 @@ def login_failure(self, *args, **kwargs):
             logging in, let us know by emailing <a href="mailto:{email}">{email}</a>.
         """).format(email='oneanddone@mozilla.com'), extra_tags='safe')
         return super(Verify, self).login_failure(*args, **kwargs)
+    
+    def login_success(self, *args, **kwargs):
+        if (not UserProfile.objects.filter(user=self.request.user).exists()):
```
bobsilverberg: @tessie There is a problem here because `self.request.user` has not been set to the logged in user yet, so this statement is generating an error. You need to use `self.user` instead. That will fix this issue, but I have come across another one that I am asking the developer of django-browserid for help with. I will update this when I have a solution.

bobsilverberg: Also, you do not need the `(` and `)` around the condition here, so you should remove them.

oneanddone/users/views.py
```diff
@@ -151,6 +151,15 @@ def login_failure(self, *args, **kwargs):
...
+            self.success_url = reverse_lazy('users.profile.create')
+            return super(Verify, self).login_success(*args, **kwargs)
+        elif (not UserProfile.objects.get(user=self.request.user).privacy_policy_accepted):
```
bobsilverberg: As above, this needs to use `self.user` and also you do not need the `(` and `)` around the condition here, so you should remove them.

bobsilverberg:
@tessie I figured out the issue. We need to override the `success_url` property, not the `login_success` method. Create your own local copy of `success_url` and use the same logic to return either `reverse_lazy('users.profile.create')`, `reverse_lazy('users.profile.update')` or `super(Verify, self).success_url`.

Give that a try and let me know how it goes.

bobsilverberg:
@tessie Is this ready to review again? Please ping me or add a comment when a PR is ready to be re-reviewed because otherwise it doesn't appear on my radar.

tessie:
@bobsilverberg .its ready for review.But the code is yet not working properly

bobsilverberg:
Thanks @tessie. Can you please explain in detail when you mean by the code is not working? Please explain what about it is not working and/or any error messages you are seeing. Just telling me it is not working does not give me enough information to be able to help you.

tessie:
@bobsilverberg .sure.I have verified that the code passes the if and elif condition.But the redirect is not working as intended.In all cases it gets redirected to the login redirect url

bobsilverberg:
Ok, thanks @tessie. I will try it out and see if I can figure out why it is doing that.

bobsilverberg:
@tessie This seems to be working fine for me. Good work!

A couple other things this branch needs are:

1. Unit tests for the new `success_url` function. This may be pretty challenging, but give it a try and look at other unit tests in the repo for inspiration.
2. We should remove the code for `dispatch` [1] in `base/views.py` as it is no longer needed with this new login code.

[1] https://github.com/mozilla/oneanddone/blob/master/oneanddone/base/views.py#L27

tessie:
@bobsilverberg .The code does not work for me even now.I cloned the repository and tried.it checks the if condtion for checking a user profile exists and if it does not enters that block.I have sucessfuly overide the success_url.I tried by overriding the login_sucess and print "self.success_url " .Its value changes accordingly.But in localhost it gets redirected to next?paramenter

bobsilverberg:
@tessie Can you describe for me the exact steps that you take when testing this? And what happens at each step? I have tried to recreate the problem you describe and I am not seeing it. Please tell me which URLs you are visiting in which order and exactly what happens.

Thanks!

tessie:
@bobsilverberg .Sure.I have checked three cases.
Case 1: The User Does not have a userprofile.
Case 2 : The User has an associated UserProfile but had not signed the privacy policy
Case 3: The User has signed the privacy policy
 
In case 1: Login from the home page and 'tasks/available url' I am getting error no matching userprofile exists.

Case 2: Signin from home page and tasks.available page and getting redirected to the same page and not userprofile update.

bobsilverberg:
Thanks @tessie. I will try to reproduce these.

bobsilverberg:
@tessie This is very strange as I do not get the same results as you. Here is what I did:

1. I found an existing user in the database which corresponds to an email address of mine.
2. In the database, I deleted the UserProfile record which corresponds to that user.
3. I tried to log in as that user, both from the home page and from the Available Tasks list.
3.1 In both cases I was redirected to the "Create Profile" page.
4. In the second test I allowed the system to create a profile for the user, so now that user has a profile.
5. In the database I edited the UserProfile record to set the `privacy_policy_accepted` flag to 0.
6. I tried to log in as that user, both from the home page and from the Available Tasks list.
6.1 In both cases I was redirected to the "Edit Profile" page so I could sign the privacy policy.

So everything seems to be working for me. How did you run your tests? Can you try duplicating the steps I have listed above and test again? I cannot seem to reproduce any of the incorrect behaviour that you are reporting.

tessie:
bobsilverberg: doest not work

tessie:
@bobsilverberg .does not work

bobsilverberg:
@tessie I have no idea what you mean by "does not work". That is not enough information. What does not work? Which of the steps above does not work for you, and what do you see instead? You need to spend more time in your written communications to make yourself clear.

tessie:
@bobsilverberg .Actually now it seems to be working for me.But I think need help in writing the test case

bobsilverberg:
Hi @tessie. It looks like we only really need some tests for the new `success_url` method in the `Verify` class. We should just check to make sure that the url returned to us is what is expected in each of the 3 cases. You should be able to use code similar to that in `CreateProfileViewTests` to create the user objects for the tests.

tessie:
@bobsilverberg .Shall be updating soon.Thanks for the feedback

tessie:
@bobsilverberg .I was attempting to fix it.Pasting th e trace

Traceback (most recent call last):
  File "/home/tess/oneanddone/oneanddone/users/tests/test_views.py", line 117, in test_redirect_url_no_profile
    eq_(self.view.success_url(self), redirect.return_value)
  File "/home/tess/oneanddone/oneanddone/users/views.py", line 157, in success_url
    if not UserProfile.objects.filter(user=self.user).exists():
AttributeError: 'Verify' object has no attribute 'user'

tessie:
@akatsoulas  would be nice if you could review my PR and help me in fixing

akatsoulas:
Hey @tessie, @bobsilverberg is doing a fantastic job and spend a good amount of time to guide you through the steps needed to fix this bug and I would argue that he is the most suitable person to help you finish this. That being said, if there is another PR without a reviewer, I would be more than happy to help.

oneanddone/users/tests/test_views.py
```diff
@@ -103,3 +103,18 @@ def test_get_object_non_existent_username(self):
 
         with self.assertRaises(Http404):
             self.view.get_object()
+
```
bobsilverberg: You need an extra blank line between classes.

oneanddone/users/tests/test_views.py
```diff
@@ -103,3 +103,18 @@ def test_get_object_non_existent_username(self):
...
+class VerifyTests(TestCase):
+    def setUp(self):
+        self.view = views.Verify()
+       
+
```
bobsilverberg: You have one blank line too many here.

oneanddone/users/tests/test_views.py
```diff
@@ -103,3 +103,18 @@ def test_get_object_non_existent_username(self):
...
+    def test_redirect_url_no_profile(self):
+        """
+        If the user has no profile, redirect  the url to create user profile.
+        """
+        with patch('oneanddone.users.views.redirect') as redirect:
+            eq_(self.view.success_url(self), redirect.return_value)
+            redirect.assert_called_with('users.profile.create')
+        
+
```
bobsilverberg: You have extra blank lines and whitespace at the end of the file.

oneanddone/users/tests/test_views.py
```diff
@@ -103,3 +103,18 @@ def test_get_object_non_existent_username(self):
...
+class VerifyTests(TestCase):
+    def setUp(self):
+        self.view = views.Verify()
+       
+
+    def test_redirect_url_no_profile(self):
+        """
+        If the user has no profile, redirect  the url to create user profile.
+        """
+        with patch('oneanddone.users.views.redirect') as redirect:
```
bobsilverberg: You don't need code like this at all. You are not testing an actual redirect, you are just testing the return value of the `success_url` property. You do need to inject a user into the view, as one is expected, which is why you are seeing that error about `'Verify' object has no attribute 'user'`. Here is an example of a test that will work for the first case. Maybe you can try creating tests for the other cases in `success_url`.

```python
    def test_success_url_no_profile(self):
        """
        If the user has no profile, return the url to create user profile.
        """
        user = UserFactory.create()
        self.view.user = user
        eq_(self.view.success_url, reverse_lazy('users.profile.create'))
```

bobsilverberg:
@tessie I reviewed your test and added some comments. Let me know if you have any questions.

bobsilverberg:
@tessie This PR is now out of date too. Please let me know if you have time to rebase it and address my last comment.

tessie:
@bobsilverberg: Can you please review now

bobsilverberg:
@tessie During our earlier conversations about this, via my comments in this pull request, we discussed adding tests for this new feature. I believe you started doing that, and I had made some suggestions about how to improve the tests, but now I don't see any tests at all. It seems that they must have been left out or deleted during your rebase of the code. Can you look back through the code and at my comments to see if you can add the tests back in?

Also, there is an odd file (https://github.com/mozilla/oneanddone/pull/237/files#diff-7c3613dba5171cb6027c67835dd3b9d4R1) which is appearing as added by this PR which should not be there.

bobsilverberg:
@tessie I added the missing unit tests and squashed all of your commits into a single commit, which landed as https://github.com/mozilla/oneanddone/commit/c2519f889371d8ee30c8def51514687f51387579. This can now be closed.

Thanks for all your hard work on this.

tessie:
@bobsilverberg  great

PR #238, opened 2014-12-29
Bug 1072341 - Invalid error displays twice

skjindal93:


bobsilverberg:
Wow, @skjindal93, it really was just as simple as that!  Great work! I'm going to merge this in.

PR #241, opened 2015-01-08
Bug 1112679 - Add local host info to the readme file

Gioyik:


rbillings:
Looks good to me

rbillings:
For reference this is for Bug 1112679: Add local host info to the readme file

bobsilverberg:
Thanks for doing this, @Gioyik. I think there is a better solution to the problem though. Instead of adding that line to the `readme.md` file, we should update the `local.py-dist` in the `settings` folder to specify:

`BROWSERID_AUDIENCES = ['http://localhost:8000', 'http://127.0.0.1:8000']`

instead of:

`BROWSERID_AUDIENCES = ['http://localhost:8000']`

Do you want to try to make that change?

Gioyik:
@bobsilverberg not problem. So, I need to revert the changes on the readme.md file?

bobsilverberg:
@Gioyik Yes, we don't need the changes to the readme.md file.

Gioyik:
@bobsilverberg done!

bobsilverberg:
This is great, @Gioyik. Thanks for doing this!

Would you be able to squash the 3 commits that are currently in this branch into a single commit, after which I will merge the changes? For some info on doing this, please see https://wiki.mozilla.org/Web_Testing/Automation/CodeReviewProcess#Squashing_commits_before_merging_a_feature_branch

Gioyik:
Commits squashed! Let me know if there's something I can do.

Thanks

bobsilverberg:
Thanks @Gioyik. Great job!

PR #244, opened 2015-01-11
Fixed Bug 1050418 - Remove ability to list and get details of Users via ...

tessie:
...the One and Done User REST API

### DIFF
--- a/oneanddone/users/tests/test_api.py
+++ b/oneanddone/users/tests/test_api.py
@@ -5,6 +5,7 @@ import random
 import json
 
 from django.contrib.auth.models import Permission
+from django.contrib.auth.models import User
 
 from rest_framework import status
 from rest_framework.test import APITestCase
@@ -102,8 +103,7 @@ class APITests(APITestCase):
         self.assert_response_status(delete_response, status.HTTP_204_NO_CONTENT)
 
         # Verify that the user has been deleted
-        get_response = self.client.get(user_uri)
-        self.assert_response_status(get_response, status.HTTP_404_NOT_FOUND)
+        eq_(User.objects.filter(username=test_user.username).count(), 0)
 
     def test_forbidden_client(self):
         """
@@ -120,24 +120,6 @@ class APITests(APITestCase):
         # Make a DELETE request
         delete_response = self.client.delete(user_uri)
         self.assert_response_status(delete_response, status.HTTP_403_FORBIDDEN)
-
-    def test_get_user_details(self):
-        """
-        Test GET details of a user with particular email for authenticated user
-        """
-        self.client.credentials(HTTP_AUTHORIZATION='Token ' + self.token.key)
-        test_user = UserFactory.create()
-        user_uri = self.uri + test_user.email + '/'
-        response = self.client.get(user_uri)
-        self.assert_response_status(response, status.HTTP_200_OK)
-
-    def test_get_user_list(self):
-        """
-        Test GET user list for authenticated user
-        """
-        header = {'HTTP_AUTHORIZATION': 'Token {}'.format(self.token)}
-        response = self.client.get(reverse('api-user'), {}, **header)
-        self.assert_response_status(response, status.HTTP_200_OK)
 
     def test_unauthenticated_client(self):
         """
--- a/oneanddone/users/views.py
+++ b/oneanddone/users/views.py
@@ -153,19 +153,18 @@ class Verify(django_browserid.views.Verify):
         return super(Verify, self).login_failure(*args, **kwargs)
 
 
-class UserDetailAPI(generics.RetrieveUpdateDestroyAPIView):
+class UserDetailAPI(generics.UpdateAPIView, generics.DestroyAPIView):
     """
-    API endpoint used to get, update and delete user data.
+    API endpoint used to update and delete user data.
     """
     lookup_field = 'email'
     queryset = User.objects.all()
     serializer_class = UserSerializer
 
 
-class UserListAPI(generics.ListCreateAPIView):
+class UserCreateAPI(generics.CreateAPIView):
     """
-    API endpoint used to get a complete list of users
-    and create a new user.
+    API endpoint used to create a new user.
     """
     queryset = User.objects.all()
     serializer_class = UserSerializer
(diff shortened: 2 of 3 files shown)


tessie:
@bobsilverberg .have removed the ability to list the users.I have not removed the ability to get detail of specific users data

tessie:
@bobsilverberg .ready to review

bobsilverberg:
@tessie This is a good start, but, as the bug says, we also want to eliminate the ability to get data for a specific user via the API. I believe that that will involve a change similar to the one you made here. I'm also going to add a couple of comments to the current PR.

oneanddone/users/views.py
```diff
@@ -162,7 +162,7 @@ class UserDetailAPI(generics.RetrieveUpdateDestroyAPIView):
     serializer_class = UserSerializer
 
 
-class UserListAPI(generics.ListCreateAPIView):
+class UserListAPI(generics.CreateAPIView):
```
bobsilverberg: As this is being changed from a class that lists users, to one that just creates users, you should update the name accordingly.

oneanddone/users/views.py
```diff
@@ -162,7 +162,7 @@ class UserDetailAPI(generics.RetrieveUpdateDestroyAPIView):
...
     """
     API endpoint used to get a complete list of users
```
bobsilverberg: As above, update the comment to reflect the fact that the behaviour of this class has changed.

bobsilverberg:
@tessie This is actually a good opportunity to get involved with the functional tests we have for One and Done, as they make use of the API. The tests live in their own repo at https://github.com/mozilla/oneanddone-tests. You should be able to find the information you need to run them there, as well as at https://developer.mozilla.org/en-US/docs/Mozilla/QA/Running_Web_QA_automated_tests.

I would suggest trying to get these tests running, and once you do, you can then use them to make sure you haven't broken the API when you make changes to the API such as in this pull request.

tessie:
@bobsilverberg .ready to review

oneanddone/users/tests/test_api.py
```diff
@@ -96,14 +97,18 @@ def test_delete_user(self):
         self.client.credentials(HTTP_AUTHORIZATION='Token ' + self.token.key)
         test_user = UserFactory.create()
         user_uri = self.uri + test_user.email + '/'
+        get_response = User.objects.get(username=test_user.username)
```
bobsilverberg: This is unnecessary as it is not being used later in the test.

oneanddone/users/tests/test_api.py
```diff
@@ -96,14 +97,18 @@ def test_delete_user(self):
...
 
         # Make a DELETE request
         delete_response = self.client.delete(user_uri)
         self.assert_response_status(delete_response, status.HTTP_204_NO_CONTENT)
 
         # Verify that the user has been deleted
-        get_response = self.client.get(user_uri)
-        self.assert_response_status(get_response, status.HTTP_404_NOT_FOUND)
+        try:
+            get_response = User.objects.get(username=test_user.username)
```
bobsilverberg: Don't use the variable name `get_response` here because it is not a response object that is being returned, it is a `User` object. Using the word `response` here is confusing.

oneanddone/users/tests/test_api.py
```diff
@@ -96,14 +97,18 @@ def test_delete_user(self):
...
+        except:
```
bobsilverberg: Using a catch-all `except` is not good practice as this will catch any exceptions, not just the one we are expecting. Instead, catch only the expected exception (which I believe is `DoesNotExist`), and if that exception is not thrown then you can fail the test. You won't need the `eq_()` below, because all you want to know is that the `DoesNotExist` exception was thrown.

You could also, alternatively, use `filter` instead of `get` and make sure that the queryset is empty. That would avoid having to catch an exception at all, so might be more readable.

oneanddone/users/tests/test_models.py
```diff
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 from django.test.utils import override_settings
+from django.shortcuts import get_object_or_404
```
bobsilverberg: Is this really needed? I don't see it used anywhere.

oneanddone/users/views.py
```diff
@@ -153,19 +153,18 @@ def login_failure(self, *args, **kwargs):
         return super(Verify, self).login_failure(*args, **kwargs)
 
 
-class UserDetailAPI(generics.RetrieveUpdateDestroyAPIView):
+class UserDetailAPI(generics.UpdateAPIView,generics.DestroyAPIView):
```
bobsilverberg: Missing a space between `generics.UpdateAPIView` and `generics.DestroyAPIView`.

bobsilverberg:
Nice work, @tessie!  Please address the comments I added and let me know when it's ready for another review.

oneanddone/users/tests/test_api.py
```diff
@@ -102,8 +103,7 @@ def test_delete_user(self):
         self.assert_response_status(delete_response, status.HTTP_204_NO_CONTENT)
 
         # Verify that the user has been deleted
-        get_response = self.client.get(user_uri)
-        self.assert_response_status(get_response, status.HTTP_404_NOT_FOUND)
+        eq_(User.objects.filter(username=test_user.username).count(), 0)
```
tessie: testing using if the querrylist is empty

oneanddone/users/views.py
```diff
@@ -153,19 +153,18 @@ def login_failure(self, *args, **kwargs):
         return super(Verify, self).login_failure(*args, **kwargs)
 
 
-class UserDetailAPI(generics.RetrieveUpdateDestroyAPIView):
+class UserDetailAPI(generics.UpdateAPIView, generics.DestroyAPIView):
     """
```
tessie: added space

tessie:
@bobsilverberg .Please review.Thank you

bobsilverberg:
Looks good @tessie. Thanks for your work. Merging!

PR #245, opened 2015-01-12
Fix Bug 1014435 - [admin] Show the duration (time taken) by the user for a task to complete

tessie:


oneanddone/tasks/models.py
```diff
@@ -34,6 +34,7 @@ def __unicode__(self):
 class Feedback(CachedModel, CreatedModifiedModel):
     attempt = models.OneToOneField('TaskAttempt')
     text = models.TextField()
+    completion_time_in_minutes = models.IntegerField(max_length=5,blank=True,null=True)
```
tessie: Added a new field.to save the time taken to complete.

tessie:
@bobsilverberg .please review

oneanddone/base/static/css/one-and-done.less
```diff
@@ -166,7 +166,7 @@ footer {
 }
 .feedback-form {
     width: 95%;
-    text-align: right;
+    text-align: left;
```
bobsilverberg: While this makes the new text input appear properly aligned, it also causes the "No Thanks" and "Submit Feedback" buttons to now appear in the lower left of the form, as opposed to the lower right, which is where they are on the current site, and where they need to stay, so this is not correct.

I would suggest that any time you are changing existing css (such as here, where you are changing an existing property from `right` to `left`), you need to be very careful and review the impact of the change on the site. This is something you could have noticed if you would have compared the feedback form currently on the site and the new feedback form produced with this change.

oneanddone/tasks/forms.py
```diff
@@ -25,9 +25,11 @@ def __init__(self, *args, **kwargs):
 
 
 class FeedbackForm(forms.ModelForm):
+    completion_time_in_minutes = forms.IntegerField(
```
bobsilverberg: I notice that if I enter a value that is too large, e.g., `999999999999999`, then an exception is thrown. I might have thought that the `max_length=5` in the model would have caused there to be an automatic validation for this, but that doesn't seem to be happening. Can you please see what can be done so that an exception will not be thrown for an invalid number?

oneanddone/tasks/forms.py
```diff
@@ -25,9 +25,11 @@ def __init__(self, *args, **kwargs):
...
+                      label= _('How many minutes did the task take to complete?'))
```
bobsilverberg: We ask for feedback both for completed attempts and also for abandoned attempts, so this label is not always going to be accurate. I suggest changing it to `How many minutes did you spend on the task?`

oneanddone/tasks/migrations/0022_auto__add_field_feedback_completion_time_in_minutes.py
```diff
@@ -0,0 +1,183 @@
+# -*- coding: utf-8 -*-
```
bobsilverberg: Because you created some migrations in an earlier pull request, the name of this file may not be accurate when it's ready to merge. I believe the next number in sequence, considering your other PR, would be `0024`.

bobsilverberg:
This is great work! Thanks @tessie. I have added a few comments to address, but overall this is very good. :+1: 

One other thing I would suggest adding would be to update the email that gets sent out to the task owner when feedback is submitted to include this new field.

Let me know when this is ready for another review by me.

oneanddone/tasks/models.py
```diff
@@ -34,6 +35,8 @@ def __unicode__(self):
 class Feedback(CachedModel, CreatedModifiedModel):
     attempt = models.OneToOneField('TaskAttempt')
     text = models.TextField()
+    How_many_minutes_did_you_spend_on_the_task = models.IntegerField(
```
tessie: This fixes the issue for me.

oneanddone/base/static/css/one-and-done.less
```diff
@@ -176,6 +176,9 @@ footer {
     .no-feedback {
         margin-right: 1rem;
     }
+    .right {
```
tessie: Now the feedback and no thanks buttons are moved to right

oneanddone/tasks/models.py
```diff
@@ -34,6 +35,8 @@ def __unicode__(self):
 class Feedback(CachedModel, CreatedModifiedModel):
     attempt = models.OneToOneField('TaskAttempt')
     text = models.TextField()
+    how_many_minutes_did_you_spend_on_the_task = models.IntegerField(
```
tessie: Changed the field name.

oneanddone/tasks/models.py
```diff
@@ -34,6 +35,8 @@ def __unicode__(self):
...
+        validators=[MinValueValidator(0), MaxValueValidator(99999)], blank=True, null=True)
```
tessie: This fixed the issue of causing exception

oneanddone/tasks/templates/tasks/emails/feedback_email.txt
```diff
@@ -5,3 +5,8 @@ The task: {{ task_name }} ({{ task_link }}) was {{ task_state }}.
 Feedback:
 
 {{ feedback }}
+
+Time spent on the task:
+
+{{ time_spent_on_task }}m
```
tessie: added to email text

tessie:
@bobsilverberg .Ready to review
Changed the model name
Added validator which now removed the exception for me

oneanddone/tasks/migrations/0023_auto__add_field_feedback_how_many_minutes_did_you_spend_on_the_task.py
```diff
@@ -0,0 +1,183 @@
+# -*- coding: utf-8 -*-
```
bobsilverberg: This one needs to be prefixed with `0024` as I mentioned in my last comment. The PR that I just merged added migrations for 0022 and 0023.

oneanddone/tasks/models.py
bobsilverberg: I think the old field name you used was better, except that it specified completed. I suggest changing this to `time_spent_in_minutes`. That is shorter and more succinct.

bobsilverberg:
Good work, @tessie. I added a couple of comments to be addressed, but other than that it looks really good.

tessie:
@bobsilverberg .time spent in minutes

bobsilverberg:
This is perfect now, thanks @tessie!

I realized that there is one other place where we should add this: to the Task Activity screen. I suggest adding a field below the date that the task was finished/abandoned/etc, indicating the amount of time spent on the task. Please only display this field if there is a value for that field, as many records will not have a value.

Please squash the commits now, as I know that everything so far in this PR is good, and then add a new commit with the changes for the Task Activity screen.

Thanks!

tessie:
@bobsilverberg .ready for review

oneanddone/tasks/templates/tasks/activity_listing.html
```diff
@@ -24,6 +24,12 @@
           <div>{{ _("Team: {team}")|fe(team=attempt.task.team.name) }}</div>
           <div>{{ _('Date {state}: {modified}')|fe(state=attempt.get_state_display(),
             modified=attempt.modified.strftime('%Y-%m-%d: %I:%M:%S %p').lower()) }}</div>
+          <div>
```
bobsilverberg: I suggest moving the `<div>` inside the `{% if` so that if we do not have a `time_spent_in_minutes` we do not even output the `<div>` to the page.

oneanddone/tasks/templates/tasks/activity_listing.html
```diff
@@ -24,6 +24,12 @@
...
+            {% if attempt.has_feedback %}
```
bobsilverberg: We need to check not only that the attempt has feedback, but also that `time_spent_in_minutes` is not `None`, because there will be old `Feedback` entries that do not have a value for `time_spent_in_minutes`.

oneanddone/tasks/templates/tasks/emails/feedback_email.txt
```diff
@@ -5,3 +5,8 @@ The task: {{ task_name }} ({{ task_link }}) was {{ task_state }}.
 Feedback:
 
 {{ feedback }}
+
+Time spent on the task:
```
bobsilverberg: Because the time spent will never be longer than 5 characters, I suggest just outputting it on the same line as this, such as:
```
Time spent on the task: {{ time_spent_on_task }}m
```

bobsilverberg:
Great work @tessie. I have added a few comments for you to address.

tessie:
@bobsilverberg .Thanks updated the pr

oneanddone/tasks/templates/tasks/activity_listing.html
```diff
@@ -24,6 +24,15 @@
           <div>{{ _("Team: {team}")|fe(team=attempt.task.team.name) }}</div>
           <div>{{ _('Date {state}: {modified}')|fe(state=attempt.get_state_display(),
             modified=attempt.modified.strftime('%Y-%m-%d: %I:%M:%S %p').lower()) }}</div>
+          {% if attempt.has_feedback %}
```
bobsilverberg: I believe that the code you wrote above for `attempt.time_spent_in_minutes` will return `None` if `attempt.has_feedback` is `False`, so I think you could just use `{% if attempt.time_spent_in_minutes %}` and not even bother checking `if attempt.has_feedback`.

bobsilverberg:
This looks good, @tessie. I added one comment to address.

Also, I think you misunderstood what I meant about squashing. Because this PR has gone on for a long time and there are lots of old commits, I wanted you to squash all the *old* commits, but then to start adding new commits after that. I didn't want you to just start squashing every time you added a commit. I want to be able to see the exact changes that were just made, so if you squash everything I cannot see that. I did explicitly say above "Please squash the commits *now*, as I know that everything so far in this PR is good, and then *add a new commit* with the changes for the Task Activity screen", but it seems like you misunderstood.

Does this make sense now?

tessie:
@bobsilverberg .thanks for review.updated the pr

bobsilverberg:
This is great, thanks @tessie!

I notice that there are now some merge conflicts because of other PRs that have been merged. Can you squash the commits and try to resolve the merge conflicts? If you have trouble doing that let me know and I can help.

tessie:
@bobsilverberg .hopefully now it is fixed

bobsilverberg:
Perfect! Thanks @tessie. Merging now!

PR #246, opened 2015-01-12
Bug 1014431 - Improve visualization for code snippets

tessie:


tessie:
@bobsilverberg .Please review.Now the code within the instruction appears in italic font

bobsilverberg:
Thanks for taking this on, @tessie, but I don't think this is quite what the bug was looking for.

Putting the code in italics won't really make it stand out much more than it already does when displayed in a code font. The bug suggested something like a lighter background or a border, which would make it stand out even more.

Do you want to give that a try?

tessie:
@bobsilverberg .I would like to try.I am not confident though.but seems challenging

tessie:
@bobsilverberg .changed the background color.Please review

bobsilverberg:
@tessie The css seems to work, but the colour you chose doesn't really show up very obviously. I suggest changing to a light yellow colour.

Also, we really shouldn't be changing values in the `sandstone/sandstone-resp.less` as that is a framework file. Can you find a way of making the same change but using our local `one-and-done.less` file?

tessie:
@bobsilverberg .Sure I would try.

tessie:
@bobsilverberg .ready for review
moved the css code to one-and-done.less

oneanddone/base/static/css/one-and-done.less
```diff
@@ -30,6 +30,11 @@ h4, .billboard h4 {
 p {
     line-height: 1.5;
 }
+pre,
+code {
```
bobsilverberg: This looks pretty good, @tessie. I suggest making a few changes:

1. Remove the italic. With the highlighting it is unnecessary.
2. Make the colour a bit lighter. It seems to be a bit too bright right now.
3. You are overriding the `pre` and `code` class from sandstone, so you should probably bring in the properties from that file too. Specifically the ones at https://github.com/mozilla/oneanddone/blob/master/oneanddone/base/static/css/sandstone/sandstone-resp.less#L229

tessie:
@bobsilverberg .ready to review

bobsilverberg:
Thanks @tessie, nice work!

I squashed the commits for you, and this landed in https://github.com/mozilla/oneanddone/commit/4ea5c2b91ebb7b5af327c2127ff4f523e44f32f1.

PR #250, opened 2015-01-16
Bug 1040682 - Feedback emails should contain link to the feedback entry

tessie:


tessie:
@bobsilverberg .please review

oneanddone/tasks/views.py
```diff
@@ -89,11 +89,15 @@ def form_valid(self, form):
         task_name = feedback.attempt.task.name
         subject = 'Feedback on %s from One and Done' % task_name
         task_link = 'http'
+        feedback_link = 'http'
         if self.request.is_secure():
             task_link += 's'
+            feedback_link += 's'
```
bobsilverberg: You are duplicating for `feedback_link` what is already being done for `task_link`. Instead of having two copies of the same code, you could just call the variable `link_prefix`, and then use that as the start of both `task_link` and `feedback_link` below.

oneanddone/tasks/templates/tasks/emails/feedback_email.txt
```diff
@@ -4,4 +4,4 @@ The task: {{ task_name }} ({{ task_link }}) was {{ task_state }}.
 
 Feedback:
 
-{{ feedback }}
+{{ feedback }}({{ feedback_link }})
```
bobsilverberg: The link ends up appearing immediately after the feedback text, which can be a bit confusing. I suggest adding a newline between the `{{ feedback }}` and `({{ feedback_link }})`.

Here is an example email with the current code:
```
---------- MESSAGE FOLLOWS ----------
Content-Type: text/plain; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: 7bit
Subject: Feedback on Try out the latest Firefox  from One and Done
From: oneanddone@mozilla.com
To: bsilverberg@mozilla.com
Date: Mon, 19 Jan 2015 21:33:07 -0000
Message-ID: <20150119213307.16734.2182@bsilverberg-09386.local>
X-Peer: 127.0.0.1

Here is your One and Done task feedback from Bob Silverberg (bsilverberg@mozilla.com):

The task: Try out the latest Firefox  (http://127.0.0.1:8000/tasks/16/) was Finished.

Feedback:

Here is some feedback.(http://127.0.0.1:8000/admin/tasks/feedback/270)
------------ END MESSAGE ------------
```

bobsilverberg:
Nice work, @tessie!  I have added a couple of comments to be addressed.

tessie:
@bobsilverberg .added link_prefix.I am unable to test send_email.ready for review

bobsilverberg:
Thanks @tessie. This works well. I have squashed your commits and landed the change in https://github.com/mozilla/oneanddone/commit/c5a68a2e25dbbaea3b92e1637789dd87a6be24c1.

For future reference, if you want to test code that sends emails, you can configure django to use a local mail server and run it locally. See this snippet from my `local.py` file:

https://gist.github.com/bobsilverberg/14b07ea8d8916296fa6b

tessie:
@bobsilverberg .Thank you

PR #252, opened 2015-01-22
Bug 1032371 - Display users who have completed task on task detail page

tessie:


tessie:
@bobsilverberg .reopening to fix issue sort users are listed  in the order most recently completed first.https://bugzilla.mozilla.org/show_bug.cgi?id=1032371

bobsilverberg:
This is good, @tessie, and works well.

I suggest you also update `test_users_who_completed_this_task` so that it verifies that the users are in the expected order.

tessie:
@bobsilverberg .updated my pr

oneanddone/tasks/tests/test_models.py
```diff
@@ -499,7 +499,7 @@ def test_abandoned_user_count(self):
     def test_all_attempts(self):
         eq_(len(self.task1.all_attempts), 11)
 
-    def test_incomplete_attempts(self):
```
tessie: found two methods with same name.at L516

bobsilverberg:
Thanks @tessie, this looks good. I took the liberty of squashing your commits for you, and then merging them, so this landed in https://github.com/mozilla/oneanddone/commit/8e7282162ad30a171cbaf1209ad36c331c8a322c

PR #253, opened 2015-01-22
Bug 1043375 - [traceback] REST API raises a 500 error - TemplateSyntaxEr...

tessie:
...ror: Encountered unknown tag 'load'.

### DIFF
--- a/oneanddone/settings/base.py
+++ b/oneanddone/settings/base.py
@@ -96,6 +96,7 @@ JINGO_EXCLUDE_APPS = [
     'registration',
     'browserid',
     'memcached',
+    'rest_framework'
 ]
 
 # Accepted locales


tessie:
@bobsilverberg .Over to review.I am getting the templates.However when i try to get data it is asking me to authenticate

bobsilverberg:
Thanks @tessie. This does resolve the 500 error. It's true that one cannot utilize the browsable api currently as it requires authentication. It looks like enabling authentication is possible [1], but I don't think we really even need it right now.

I would suggest we can implement this, which fixes bug 1043375, for now, and later, if we actually want to make the browsable api usable, we can look into adding login/authentication to it.

[1]  https://groups.google.com/forum/#!topic/django-rest-framework/80nLymebYRg

PR #262, opened 2015-02-15
Bug 1014451 - Add task rating for easier feedback analysis

tessie:
Bug 1014451 - Add task rating for easier feedback analysis


tessie:
@bobsilverberg .waiting for review

bobsilverberg:
Thanks for taking this on, @tessie. This is a good start.

I don't think the current implementation is very attractive though. One issue is that the images you have are different sizes, and you can see this by the differing quality of the icons on the page. I suggest you look at Henrik's suggestion at https://bugzilla.mozilla.org/show_bug.cgi?id=1014451#c0, and maybe you can borrow the code and technique used by input.mozilla.org.

oneanddone/tasks/forms.py
```diff
@@ -25,12 +27,16 @@ def __init__(self, *args, **kwargs):
 
 
 class FeedbackForm(forms.ModelForm):
+    CHOICES = ((1, mark_safe(_('<img src="/static/icons/happy.png" alt="approve" width= "87" height = "87" title="approve">'))),
```
bobsilverberg: `CHOICES` is not a very descriptive name for this variable. There could be other items on the form that require choices. I suggest you use the name of the field (in this case `rating`) as part of the variable name.

oneanddone/tasks/models.py
```diff
@@ -37,6 +37,7 @@ class Feedback(CachedModel, CreatedModifiedModel):
     text = models.TextField()
     time_spent_in_minutes = models.IntegerField(
         validators=[MinValueValidator(0), MaxValueValidator(99999)], blank=True, null=True)
+    rating = models.IntegerField(default = 0)
```
bobsilverberg: +1 to this. Good idea to make the default `0` so that any that haven't been rated won't affect the total.

oneanddone/tasks/models.py
```diff
@@ -367,7 +374,7 @@ class Task(CachedModel, CreatedModifiedModel, CreatedByModel):
     instructions = models.TextField()
     is_draft = models.BooleanField(verbose_name='draft')
     is_invalid = models.BooleanField(verbose_name='invalid')
-    name = models.CharField(max_length=255, verbose_name='title', unique=True)
```
bobsilverberg: I assume this was a mistake. We don't want to remove the `unique` argument from this `CharField`.

oneanddone/tasks/models.py
```diff
@@ -517,6 +524,14 @@ def users_who_completed_this_task(self):
                 users.append(user)
         return users
 
+    @property
+    def total_votes(self):
```
bobsilverberg: Sometimes this is called `rating` and sometimes you use the term `votes`, which is confusing. I think `rating` makes more sense, so I suggest you use the term `rating` everywhere and do away with `votes`.

oneanddone/tasks/models.py
```diff
@@ -517,6 +524,14 @@ def users_who_completed_this_task(self):
...
+       rating= 0
+       for attempt in self.all_attempts:
```
bobsilverberg: You can probably achieve this in less code using django's `aggregate` and `Sum`. Check out https://docs.djangoproject.com/en/1.4/topics/db/aggregation/

oneanddone/tasks/models.py
```diff
@@ -517,6 +524,14 @@ def users_who_completed_this_task(self):
...
+           rating+= attempt.task_rating
+       print rating,"rating"
```
bobsilverberg: It looks like you accidentally left in a print statement here.

oneanddone/tasks/templates/tasks/detail.html
```diff
@@ -19,6 +19,9 @@
   <div class="row">
     <div>{{ task.short_description }}</div>
   </div>
+  <div class="row">
+    <div>{{ _('Votes') }}: {{ task.total_votes }}</div>
```
bobsilverberg: Change `Votes` to `Rating`.

oneanddone/tasks/templates/tasks/detail.html
```diff
@@ -19,6 +19,9 @@
...
+    <div>
```
bobsilverberg: This `<div>` is not properly indented.

tessie:
@bobsilverberg .Yes.That true.So I was wondering How can I use font character for smileys

bobsilverberg:
@tessie Look at the source code for the feedback page on input.mozilla.org. It looks like they are using `<svg>` combined with some css to achieve what they are doing, which seems like a pretty decent idea to me.

bobsilverberg:
@tessie I have finished my review and added a few more comments to address, in addition to making the form look better.

tessie:
@bobsilverberg .Thanks for the feedback.Also would  Just like input.mozilla.org do you think its beter to ask the rating then according redirect to the present feedback page

tessie:
Also I think need to add sort the list view.So how should the design be changed.

bobsilverberg:
@tessie I don't understand your last comment. I think that we should just keep it as a single form, as you've done, but change the images to be buttons where only one can be selected. Use the code from input.mozilla.org as an example of how the buttons should work, or you could just use images as you have done, but find images that are the same size. Maybe @rbillings has some ideas about how it should look. If you can show her a screen shot of what you have now maybe she can make some suggestions.

tessie:
@bobsilverberg .have updated the PR

tessie:
@bobsilverberg .I have updated the smiley similar to what input.mozilla.org.As from the mal rebecca suggested i am not displaying the rating in the detail page.Let us try to fix it as another bug

bobsilverberg:
@tessie This PR needs to be rebased against the latest master to bring it up to date, and also you'll need to remove the South migration and replace it with a Django migration.

rbillings:
@tessie this still needs a rebase & Django migration change, are you up for that?

tessie:
@rbillings .Yea I am in.I thought I already did that

tessie:
@rbillings  @bobsilverberg  an issue with peep

File "./bin/peep.py", line 902, in <module>
    exit(main())
  File "./bin/peep.py", line 874, in main
    return commands[argv[1]](argv[2:])
  File "./bin/peep.py", line 834, in peep_install
    for path in req_paths))
  File "./bin/peep.py", line 834, in <genexpr>
    for path in req_paths))
  File "./bin/peep.py", line 795, in downloaded_reqs_from_path
    finder = package_finder(argv)
  File "./bin/peep.py", line 335, in package_finder
    return PackageFinder(index_urls=index_urls, **kwargs)
TypeError: __init__() got an unexpected keyword argument 'use_wheel

rbillings:
You might find help here: https://github.com/erikrose/peep
I would begin by creating a new virtualenv. Hope that helps!

rbillings:
@tessie - are you going to be able to finish this soon?

tessie:
@rbillings  I dont think So.It is free for taking up

davehunt:
Thanks @tessie. Perhaps @karlht or @rbillings might be interested in picking this up.

rbillings:
I'll close this for now, but it'll be saved in this history if anyone wants to continue working on it.

PR #269, opened 2015-04-30
Ditch playdoh and update to Django1.7

bobsilverberg:
This is a huge PR, I know, and I apologize for that, but it pretty much had to be done in one go.

The site is working for me locally, and the unit tests all pass. I also have deployed this branch to dev [1] and it too seems to be happy.

This is a very important update that needs to be done, and I'm not extremely confident in everything I've done here, as I was often just hacking away until it worked. Some of this stuff is very new to me, so it's entirely possible I made some mistakes which should be corrected before merging and deploying this to stage and prod. Therefore a thorough review is needed and will be appreciated.

@Osmose Would you be able to find time to look this over? Pretty please?
@glogiotatidis and/or @akatsoulas Your input/review would also be appreciated as this is such a huge change.

Thanks in advance to all of you! :)

[1] https://oneanddone-dev.paas.allizom.org/


bobsilverberg:
Just FYI, the number of actual changes isn't that huge. This is mostly deleting a ton of stuff and updating infrastructure. The changes required to the apps themselves are quite minor.

Osmose:
I'm having trouble commenting directly on the diff mainly because it's so large, so I'll just comment directly on the PR.

Osmose:
The problem with putting monkeypatches in the `manage.py` file is that they don't take effect when the site is run via WSGI (or rather, I would expect them not to, but maybe `manage.py` just happens to get loaded anyway?). That's why people put them in the main `urls.py` or, with Django 1.8, maybe the AppConfig for the base app?

Osmose:
In the monkeypatches, you call `load_helpers` for jingo, but IIRC jingo automatically runs that right before it renders the first template you ask it to, and I don't see it in funfactory's existing monkeypatches. Why'd you add that in?

Osmose:
`EXTRA_APPS` seems like an unnecessary setting.

Osmose:
You should probably be setting `HMAC_KEYS` from the environment as well.

Osmose:
django-decouple seems awesome!

Osmose:
Why keep the old south migrations around? Or is that just temporary?

Osmose:
Oh, I see you monkeypatched in `urls.py` as well. Maybe remove the patches from `manage.py` then?

Osmose:
I've looked over the important bits and everything seems to be in order. That dev is happy is promising as well.

The README could use some updates on how to run this locally, as I'm a little lost on what to do about the missing environment variables that django-decouple is complaining about, but that's not a blocker IMO.

Nice work! :D

bobsilverberg:
@Osmose Thanks for the review! I am unable to reply to your comments individually via the UI as they are added to the discussion and not the diff (I understand why), so I will attempt to reply to them here:

First overall response: Much of the work I did was trying to deduce what was necessary from looking at playdoh, sugardough and also willkg's PR to remove playdoh from fjord, so I made a bunch of assumptions without really understanding what the code does in all cases. So many of the answers to your questions of "why did you" are "I don't know". Having said that I did go through the "why did you" and tried not doing those things, and in many cases everything seems fine. Some examples include:

- I removed the call to `monkeypatches.patch()` from `manage.py`
- I removed the call to `load_helpers` from `monkeypatches.py`

Other changes to address your comments:

- I removed `EXTRA_APPS` from `settings/base.py`
- I am getting an `HMAC_KEY` from config, in a similar way I was getting the `SECRET_KEY`

Regarding the South migrations, that's a good question. I guess I was afraid of deleting them, but they'll remain in the repo's history so I guess I should just get rid of them entirely.

I added a commit which includes all of these changes and the site still seems to be happy locally and on Stackato dev. :)

You are correct that the readme needs to be updated as well, so I will undertake that in a separate PR. 

Could I interest you in a final r? on this? :cake:

Osmose:
Last commit looks good, I'd say r+ if you're confident in the functionality of the dev instance.

bobsilverberg:
Thanks @Osmose. I added a new commit today which moves more of the settings to use decouple and eliminates most of the custom code from settings/stackato.py. No need to review that though, unless you really want to.

@akatsoulas Is going to take a look on Monday as well, so once that's done I will go ahead and merge this and then try to get it up on staging as well. 

Thanks for all your help!

glogiotatidis:
So django 1.8 supports different templating engines (although not as good as it sounds, it's not just a config change) and we upgraded sugardough recently to django 1.8 and django-jinja. This hopefully will remove some jingo craziness and code that only mozilla uses.  Also Django 1.8 is a LTS release. 

I understand that these are too many changes so upgrading gradually makes sense. I just wanted to express my preference to django1.8 and django-jinja :)

+1 for using decouple to move configuration to env variables!

nice job! :)

akatsoulas:
I don't feel that I have anything to add here, since @Osmose has everything covered with the review and @glogiotatidis already expressed an interest in Django 1.8 :)

I was able to run this locally and play around with some of the functionality without any problems.

Really nice work!

bobsilverberg:
Thanks @glogiotatidis and @akatsoulas!  @glogiotatidis I definitely plan a follow up to move to Django 1.8 and django-jinja, but that will happen in a separate PR. It also might not happen until some other Q2 goals have been met. ;)

This code seems to be running well on both dev and staging, so I'm going to merge it, and hopefully it will make it to production soon.

bobsilverberg:
@Osmose @akatsoulas @glogiotatidis This was pushed to prod today with no issues! One and Done is finally done with Playdoh (pun intended) and on a modern version of Django.

Thanks to all of you for your help!

PR #270, opened 2015-05-04
Bug 1148046 - Allow users to start multiple tasks

bobsilverberg:
@glogiotatidis or @akatsoulas r? please

### DIFF
--- a/oneanddone/base/templates/base/dashboard.html
+++ b/oneanddone/base/templates/base/dashboard.html
@@ -23,9 +23,7 @@
 
   {% if not page == 'ProfileDetails' %}
       <nav class="actions-container">
-      {% if not user.attempts_in_progress %}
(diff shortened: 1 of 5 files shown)


akatsoulas:
@bobsilverberg If you think that it's of any value, you can add a test to showcase that a user can pick up multiple tasks. Other than that, I tested this locally and it's working fine! r+(wc?) :+1:

bobsilverberg:
Thanks @akatsoulas. There actually already is a test that has a user taking multiple tasks, which is interesting because it wasn't actually allowed before. The thing is that there was nothing in the model to prevent it. All the code was in the views.

Thanks for the review. I am going to hold off merging this until I can address the problem that currently exists getting the Django 1.7 upgrade into production (which is https://bugzilla.mozilla.org/show_bug.cgi?id=1161644 in case you are interested).

bobsilverberg:
I'm going to merge this now so I can submit a couple of others for review. Unfortunately we are still unable to deploy the latest code to prod, but I hope that may be fixed soon.

PR #271, opened 2015-05-04
Removed references to QA

rbillings:
As part of bug 1145286, removing text that refers to One and Done as a QA specific tool.


stephendonner:
r+ - thanks for taking care of this!

PR #274, opened 2015-05-07
Update readme and local.py-dist to support changes for Django 1.7 upgrade

bobsilverberg:
Also include .env-dist

Changes to the readme to support the new structure of the app

@glogiotatidis or @akatsoulas r?

### DIFF
--- a/.env-dist
+++ b/.env-dist
@@ -1,0 +1,7 @@
+DJANGO_DEBUG=True
+ALLOWED_HOSTS=localhost, 127.0.0.1
+DJANGO_SECRET_KEY=some secret
+DJANGO_HMAC_KEY=some secret
+DATABASE_URL=postgres://user:password@localhost/oneanddone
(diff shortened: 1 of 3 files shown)


.env-dist
```diff
@@ -0,0 +1,7 @@
+DJANGO_DEBUG=True
+ALLOWED_HOSTS=localhost, 127.0.0.1
+DJANGO_SECRET_KEY=some secret
+DJANGO_HMAC_KEY=some secret
+DATABASE_URL=mysql://root:@localhost:3306/oneanddone
```
akatsoulas: Minor but maybe you could add a password here, like `mysql://root:password@localhost:3306/oneanddone` to better indicate to the users where they can input their mysql credential, if any. Also this may need some explaining in the README file.

akatsoulas:
I was able to setup my local environment successfully, following the instructions in the README file. r+wc :coffee:

bobsilverberg:
Thanks @akatsoulas. I will make that suggested change and add a comment to the readme.

PR #276, opened 2015-06-17
More timely tasks on the front page (hack!)

karlht:
This prioritizes things that have a limited time scope and breaks ties in (priority, difficulty) in favour of the more recent item rather than the less recent.


karlht:
I realize that the ongoing search/front page work may make this simple hack obsolete, but I promised Stuart I'd submit the pull request.  Feel free to ignore or reject it if it's not useful.

bobsilverberg:
Hey @karlht, thanks for the pull request. It looks good to me, but I want to test it before merging, which I may not have time to do until next week.

Unfortunately, even after is it merged it won't make it to production for awhile as we have an issue currently that is preventing us from pushing new code to prod.

bobsilverberg:
@karlht Sorry to take so long to review this. There is a problem with the current approach because many tasks do not have an end date at all, and those are showing up before any tasks that do have an end date. I assume you wanted any tasks with end dates to show up before any without, and those with end dates should be sorted with the ones that will expire earliest at the top. Unfortunately, because of blank end dates, the current code you have submitted doesn't do that. I think it's worthwhile working on this some more, though, to see if you can come up with a solution that does work.

Thanks for the PR!

bobsilverberg:
@karlht Do you want to give this a try and see if you can come up with a way of making it work with blank end dates?

karlht:
I may have some time for this toward mid-to-late October.  I did some preliminary looking into it just after you reviewed it, but didn't find anything straightforward.  I still think it's a good idea, so I'll have a look, but I'm solidly booked for the next 2-3 weeks.

davehunt:
Are you still interested in working on this @karlht?

rbillings:
@karlht - unless you have some time/energy for this, perhaps we should close this?

karlht:
Sure, go ahead and close it.  If I decide it desperately needs doing, I'll re-open.

rbillings:
Closing with the option of re-opening in the future.

PR #277, opened 2015-06-29
Bug 1030984 - Create tasks that require sign off

bobsilverberg:
This is a pretty big patch that addresses all of the requirements for the bug mentioned above. @rbillings has already verified the behaviour, so it just needs a code review prior to merging.

@glogiotatidis | @akatsoulas r? please and thank you!


glogiotatidis:
@bobsilverberg tests fail, can you please take a look?

bobsilverberg:
Sorry @glogiotatidis. I guess in my haste to get a PR up for review I neglected the tests. I will fix them, and probably add some new ones for the new functionality. I'll ping again when it's ready to review.

bobsilverberg:
@glogiotatidis I fixed the broken tests and added some more for the new functionality. r?

glogiotatidis:
```Your models have changes that are not yet reflected in a migration, and so won't be applied.```
another `./manage.py makemigrations` is required

File: oneanddone/tasks/templates/tasks/form.html [hunk 1]
```diff
@@ -94,6 +94,11 @@
     {{ form_field(task_form['why_this_matters'], 'why-this-matters') }}
     {{ form_field(task_form['prerequisites']) }}
     {{ form_field(task_form['instructions']) }}
+    <label for="id_must_be_verified">
+      {{ task_form['must_be_verified'] }}
+      {{ _('Verified task') }}
```
glogiotatidis: Maybe `Must be verified` instead of `verified task`?

oneanddone/tasks/models.py
```diff
@@ -92,9 +92,27 @@ def has_feedback(self):
...
+    def needs_verification(self):
+        return self.task.must_be_verified and self.state == self.STARTED
+
+    @property
     def next_task(self):
         return self.task.next_task
 
+    @property
+    def verification_status(self):
+        if (not len(self.communication_set.all()) or
```
glogiotatidis: `not self.communication_set.exists()` is cheaper. `len(..all())` will force all results from db to be loaded and then counted in python.

oneanddone/tasks/templates/tasks/detail.html
```diff
@@ -164,5 +181,32 @@
...
+        </fieldset>
+      </div>
+    {% endif %}
+    <div class="row">
+      <div>
+        <fieldset class="verify-task-fieldset">
+          <legend>{{ _('Task Verification') }}</legend>
+          <div>
+            {{ _('Administrators will verify your work to complete this task.
+                Follow the instructions and click the button to submit your work.') }}
```
glogiotatidis: Use `{% trans %}` blocks for multiline strings. Otherwise I believe tower will extract a string with many whitespaces between `this task.` and `Follow`

oneanddone/tasks/views.py
```diff
@@ -350,19 +351,125 @@ def post(self, *args, **kwargs):
...
+        communication = form.save(commit=False)
+        communication.attempt = attempt
+        communication.type = TaskAttemptCommunication.ADMIN
+        communication.creator = self.request.user
+        communication.save()
+
+        # Send email to attempt user
+        if attempt.user.profile.consent_to_email:
+            task_name = attempt.task.name
+            subject = 'One and Done task verification update for %s' % task_name
```
glogiotatidis: Ha! The common problem with emails and localization. I had to deal with this on mozillians too. I believe the path of least resistance is what you've done here, i.e. use english. Ideally I guess we'd have a language dropdown in user preferences and then use this locale to send emails. But this is way too complicated for the needs of this app.

oneanddone/tasks/views.py
```diff
@@ -350,19 +351,125 @@ def post(self, *args, **kwargs):
...
+            link_prefix = 'http'
+            if self.request.is_secure():
+                link_prefix += 's'
+            task_link = link_prefix + '://%s%s' % (
+                self.request.get_host(),
```
glogiotatidis: Traditionally we've been using a `SITE_URL` settings entry, which can simplify the code here and in other places that you need the full url. This is fine too of course.

`SITE_URL=http://oneanddone.example.com`

oneanddone/tasks/views.py
```diff
@@ -350,19 +351,125 @@ def post(self, *args, **kwargs):
...
+                'content': communication.content})
+
+            # Manually replace quotes and double-quotes as these get
+            # escaped by the template and this makes the message look bad.
+            filtered_message = message.replace('&#34;', '"').replace('&#39;', "'")
+
+            send_mail(
+                subject,
+                filtered_message,
+                'oneanddone@mozilla.com',
```
glogiotatidis: maybe you want to move this to settings.

oneanddone/tasks/views.py
```diff
@@ -350,19 +351,125 @@ def post(self, *args, **kwargs):
...
+
+    def post(self, request, *args, **kwargs):
+        self.object = self.get_object()
+        return super(SubmitAdminVerificationResponseView, self).post(request, *args, **kwargs)
+
+    def form_valid(self, form):
+
+        attempt = self.get_object()
+
+        communication = form.save(commit=False)
```
glogiotatidis: It could make sense to move the email sending code in the `SubmitVerifiedTaskForm` to keep it DRY and more tight. So when you save the form emails go out etc.

glogiotatidis:
My experience from reps.m.o which sends a bunch of emails to mentors and reps is that users will always reply to your automated emails. Especially when you proxy communication of real persons and you don't just send automated emails. I'd suggest for both privacy reasons (personal communication might reach oneanddone@m.com) and for user awareness to use an invalid sender address for all the emails originating from the website.

glogiotatidis:
Heads up from the bug reqs
`- need to add ability to *search* for and view task attempts in the django admin`
`- record of submitted work and admin replies are stored in the django admin`
I didn't see those

glogiotatidis:
The code looks good r+wc! This is a big feature, good job! :sunglasses:

oneanddone/tasks/templates/tasks/form.html
bobsilverberg: Good suggestion. I'll change that.

oneanddone/tasks/models.py
bobsilverberg: Thanks. Good advice. :)

oneanddone/tasks/templates/tasks/detail.html
bobsilverberg: Thanks @glogiotatidis. Noted!

oneanddone/tasks/views.py
bobsilverberg: Good point. As you've no doubt noticed I often forget about i18n. Glad to hear that you think this is acceptable. :)

oneanddone/tasks/views.py
bobsilverberg: @glogiotatidis Would you use code similar to the above to set `SITE_URL` in settings? I'm not sure how we'd get this to work across sites if the setting is hardcoded. Or perhaps you use a different settings file per deployment? I'm currently working on moving the app off of Stackato, and I'm guessing that deployment-specific setting will reside somewhere, but for now I'm not sure how that will work.

oneanddone/tasks/views.py
bobsilverberg: As mentioned above, I need to figure out how site-specific settings are going to work with the new platform for the site. Stackato uses environment variables, but I think we'll need to move to something else. I will keep this in mind when I do that work.

oneanddone/tasks/views.py
bobsilverberg: Good idea. I've just done a major refactor where I moved all of the email code into a form helper method so that each view can just call `send_email` on the form object and pass it the necessary data. Thanks for the suggestion!

bobsilverberg:
@glogiotatidis Good point about the email address. AFAIK the address oneanddone@mozilla.com does not exist as a valid email address, but maybe I should change it to something like do_not_reply@mozilla.com or some such to avoid issues such as you've described in https://github.com/mozilla/oneanddone/pull/277#issuecomment-118782322.

@rbillings What do you think?

bobsilverberg:
@glogiotatidis Thanks for noticing those differences between the solution and the bug requirements. I implemented front end versions of those (although search is currently missing) so do not plan on providing a less useful admin interface to them. I believe the requirements assumed the only access to attempts would be via the django admin, but in conversations with @rbillings we realized that was less than ideal. I think, as this is pretty much complete now, I'll leave searching attempts as a future enhancement, if users ask for it.

bobsilverberg:
@glogiotatidis Thanks for the detailed review! I made your suggested changes, and also undertook a refactor of the email code. I know you said r+wc, but if you have time to look over the new commit that I added with the changes and the refactor that would be great. If you don't have time that's fine too - just let me know. Thanks!

rbillings:
In fact the oneanddone@mozilla.com email address does exist. I am a delegated user for the address. It receives email, but I've never once received a direct reply from a user at that address. I don't think it needs to be changed, but wouldn't object too much if you made it impossible to reply to emails. I'd like to keep the specific address in case we wanted to advertise contact/issues/bugs for the site- even though we haven't done that so far.

bobsilverberg:
Thanks @rbillings. If you think it would be interested to see if anyone does try to reply to any of these automated emails, then we can keep it as is, and you'll see any responses. If you'd rather I change it to an email address that won't work, let me know, and perhaps suggest an address to use.

rbillings:
@bobsilverberg - No one has responded in over a year, so it's fine to keep the same email address, no change needed.

oneanddone/tasks/views.py
glogiotatidis: I'd use environment variables. In `masterfirefoxos` project for example - which is based of sugardough - we only have `settings.base` and no `settings.local`. With the use of `decouple` python library we read values set in environment variables.

https://github.com/mozilla/masterfirefoxos/blob/master/masterfirefoxos/settings/base.py

bobsilverberg: Thanks @glogiotatidis. That's what I was doing with a number of settings already, but now that we're moving to a Python cluster I'm told we cannot use environment variables anymore.

glogiotatidis: In this case i'm using a local.py in the server. :/

PR #279, opened 2015-09-02
Use django-maintenancemode to enable maintenance mode

bobsilverberg:
This uses `django-maintenancemode` to allow us to put the app into maintenance mode. This is needed to we can take a database backup and then restore to Heroku and wait for DNS to propagate.


bobsilverberg:
This is no longer needed.

PR #298, opened 2015-10-13
Bug 1214322 - Update Django Rest Framework to v3.2

bobsilverberg:
Note that this removes the ability to update users via the API, which was never needed nor used.

@glogiotatidis or @akatsoulas r?

### DIFF
--- a/oneanddone/tasks/serializers.py
+++ b/oneanddone/tasks/serializers.py
@@ -3,12 +3,18 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 from rest_framework import serializers
 
-from oneanddone.tasks.models import Task, TaskKeyword, TaskAttempt
+from django.contrib.auth.models import User
+
+from oneanddone.tasks.models import (Task, TaskAttempt, TaskKeyword,
+                                     TaskProject, TaskTeam, TaskType)
 
 
 class TaskAttemptSerializer(serializers.ModelSerializer):
 
-    user = serializers.SlugRelatedField(many=False, slug_field='email')
+    user = serializers.SlugRelatedField(
+        many=False,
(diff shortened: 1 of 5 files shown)


akatsoulas:
@bobsilverberg sure, I 'll review it no later than tomorrow

bobsilverberg:
Thanks @akatsoulas !

oneanddone/tasks/serializers.py
```diff
@@ -3,12 +3,18 @@
...
+                                     TaskProject, TaskTeam, TaskType)
 
 
 class TaskAttemptSerializer(serializers.ModelSerializer):
 
-    user = serializers.SlugRelatedField(many=False, slug_field='email')
+    user = serializers.SlugRelatedField(
+        many=False,
+        queryset=User.objects.all(),
+        slug_field='email')
```
akatsoulas: Since this seems to be a read-write field, have you consider using `unique=True`?

bobsilverberg: I don't think I understand what `unique=True` would do here. A given `user` could have multiple `TaskAttempt`s, so is `unique=True` accurate?

akatsoulas: Indeed for this serializer you don't need this, ignore the comment above.

akatsoulas: `unique=True` in your model will make sure that you have unique emails. If there are two identical emails in your db, probably the serializer above will produce an error.

akatsoulas:
Code looks good but I haven't tested this locally. r+

bobsilverberg:
Thanks for the review, @akatsoulas. Based on our conversation above and your r+ I think this is ready to merge. I have another one related to the API which I will be issuing shortly.

PR #300, opened 2015-10-19
Bug 1050417 - Remove One and Done User REST API

bobsilverberg:


bobsilverberg:
@akatsoulas For a number of reasons we have decided to remove the Users API, but we're going to keep the Task API for now.

r?

oneanddone/users/views.py
```diff
@@ -1,27 +1,25 @@
...
 from django.http import Http404
 from django.shortcuts import redirect
 from django.views import generic
 
-from rest_framework import generics
 from braces.views import LoginRequiredMixin
 import django_browserid.views
-from tower import ugettext as _
 from random import randint
 import re
```
akatsoulas: Personally, I prefer all the `import` statements before `from`.

akatsoulas:
Apart from the minor comment, everything looks good here. r+ :rocket:

bobsilverberg:
Thanks for the review @akatsoulas!

PR #302, opened 2015-11-21
Bug 1225172 - Create Recent Contributors module

bobsilverberg:


bobsilverberg:
@rbillings I have deployed these changes to a review app at https://oneanddone-dev-pr-302.herokuapp.com/. You should be able to log into the app with your mozilla.com email address. The database is mostly empty, so you'll need to add a task or two and then can check out this new functionality. It's not necessarily as pretty as we might like it to be in the future, but I think it's acceptable and functional. I can certainly tweak the appearance but I'd also like to get this feature landed before taking on the bigger task of totally redesigning the home page.

Please let me know what you think once you are able to check it out.

bobsilverberg:
Closing this to re-open to test changes I've made to the review app config

PR #307, opened 2015-11-25
Add functional tests and run them via Travis

bobsilverberg:
This adds some functional tests to the oneanddone repo and allows them to be run by Travis. There is definitely room for improvement here, but I wanted to try to get this initial version landed and then iterate on improvements.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,59 @@
+# This indicates to Travis that we will not use or need sudo
+# so that we can benefit from and use the cache->directories
+# directive.
+sudo: no
+
+env: PIP_DOWNLOAD_CACHE="pip_cache"
+cache:
+    directories:
+        - pip_cache
+
 language: python
+
 python:
-  - "2.7"
-before_script:
-  - psql -c 'create database oneanddone;' -U postgres
-  - cp oneanddone/settings/local.py-travis oneanddone/settings/local.py
-install:
-  - pip install -r requirements.txt
+    - 2.7
+
+addons:
+  artifacts:
+    paths:
+    - results
+    permissions: public-read
+    bucket: oneanddone-tests
+
 env:
-  - DJANGO_DEBUG=False DJANGO_SECRET_KEY='FOO' DJANGO_HMAC_KEY='FOO' ALLOWED_HOSTS=localhost, DATABASE_URL=postgres://postgres:@localhost/oneanddone
-script: python manage.py test
+  global:
+    - ALLOWED_HOSTS='localhost, 127.0.0.1'
+    - DATABASE_URL=postgres://postgres:@localhost/oneanddone
+    - DISPLAY=:99.0
+    - DJANGO_HMAC_KEY='FOO'
+    - DJANGO_LIVE_TEST_SERVER_ADDRESS=localhost:8081
+    - DJANGO_SECRET_KEY='FOO'
+    - DJANGO_SETTINGS_MODULE=oneanddone.settings
+    - secure: "LYA4hvs1IoWqzX3/edq8ZNp/yMiaZ52xYNTnrvnSf3FFca3jAkTclyhuZE18AGXeNYAuk5iakWkdSm7DLsjZPJCvA2PqKMnV2DFGKSS3N+z3q+FXYl0seAOOg8nFB0HNOd3GbcgbB4kbENZpaGxPjCY0DHKoeyq4F7mItjTLRL8="
+    - secure: "FfX23m67dlVv+j0amt1tzLjyRMaV/1RV4QAqkdoELGgaUlCpmBaL0X4JRzhIMFKaPrbJMHcGq0cbIYSlYGlOPcp98MvTPD0o/QvSxB1HxbNbNVqGH4airG6BYNZehh33HmE4+Mi7xxhBHZhXRvjBhHLX/0/3+N3BV+blvMH7LC4="
+    - secure: "cxofDH+XU47UFIusirAD0ndo62fY/uIi/fUcASv9qqiJbowevrqLZdDiCj/bN433ZLlonLNmidItnavoIQ5DgRsoL5Vn4Z0utlgBmlDC5P54GWVjYuGqUmwOzXuZtZT3Kep5x7Jb8gDjyOAuwW2DA1Kni+oGCNKzEWvBcaAJQ7k="
+    - secure: "ICqGl+qB4W+PknSlrtqgbVCfaCkvosPJuol3RPC5dQ+2FCylFLMA1jmpEENcckmvsTaQGOg7WXhLEH5YARd0fzBRNStgDv43li/1Q7/yQwPMQNM3MI0X/V31PaWGeXd55s0P5ZvM7hUUiFNhm4fNgyBTgQrr/piyH12SdLK71KA="
+
+install:
+  - "pip install -r requirements.txt"
+  - "pip install -r requirements/test.txt"
+  - "npm install"
+  - "psql -c 'create database oneanddone;' -U postgres"
+
+before_script:
+  - "sh -e /etc/init.d/xvfb start"
+  - "cp oneanddone/settings/local.py-travis oneanddone/settings/local.py"
+  - "python manage.py migrate --noinput"
+  - "python manage.py collectstatic --noinput"
+
+script:
+  - "python -m smtpd -n -c DebuggingServer localhost:1025 &"
+  - "python manage.py test oneanddone/tasks/ oneanddone/users/"
+  - "py.test -v --driver Firefox --html results/results.html oneanddone/tests/functional/"
+
+notifications:
+    irc:
+        channels:
+            - "irc.mozilla.org#mozwebqa"
+        on_success: change
+        on_failure: change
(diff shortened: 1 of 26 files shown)


bobsilverberg:
@davehunt and @peterbe r? please, if you can find the time. :)

.travis.yml
```diff
@@ -1,11 +1,37 @@
...
-  - cp oneanddone/settings/local.py-travis oneanddone/settings/local.py
+- "2.7"
+node_js:
+- "0.10"
+addons:
+  artifacts:
+    paths:
+    - results
+    permissions: public-read
+    bucket: fireplace-tests
```
davehunt: Did you mean to leave this in? While artifacts may be useful I think we should have a more appropriate bucket name.

.travis.yml
```diff
@@ -1,11 +1,37 @@
...
+before_install:
+  - "export DISPLAY=:99.0"
```
davehunt: You could put this with the other global environment variables below.

.travis.yml
```diff
@@ -1,11 +1,37 @@
...
+  - "sh -e /etc/init.d/xvfb start"
 install:
-  - pip install -r requirements.txt
+  - "sudo pip install -r requirements.txt"
+  - "sudo pip install -r oneanddone/functional_tests/requirements.txt"
```
davehunt: I would suggest a `requirements` directory with `dev.txt` and `test.txt` or similar.

oneanddone/functional_tests/conftest.py
```diff
@@ -0,0 +1,42 @@
...
+        'username': persona_test_user['email'].split('@')[0],
+        'url': 'http://www.mozilla.org/'
+    }
+
+
+@pytest.fixture(scope='session')
+def base_url(base_url, live_server):
+    if base_url is None:
+        return live_server.url
+    return base_url
```
davehunt: I know we already amused ourselves with the various ways to write this, but here's another that you might actually like:

```python
def base_url(base_url, live_server):
    return base_url or live_server.url
```

Also, did you try without the `scope='session'`?

oneanddone/functional_tests/requirements.txt
```diff
@@ -0,0 +1,4 @@
+bidpom
+pytest-selenium
+requests==2.8.1
+-e git+https://github.com/bobsilverberg/pytest-django.git@6d8d45c4fd1eb58b553045dd844e35885fcf67e6#egg=pytest_django
```
davehunt: Can we use a released version of pytest-django?

oneanddone/functional_tests/conftest.py
```diff
@@ -0,0 +1,42 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
davehunt: I think I'd prefer a path of `tests/functional` as it works well when we have different suites of tests.

oneanddone/functional_tests/pages/base.py
```diff
@@ -0,0 +1,60 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
davehunt: Are these tests simply trnasplanted from oneanddone-tests? I'm just wondering how detailed you want them reviewed.

pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+DJANGO_SETTINGS_MODULE=oneanddone.settings
+addopts = --reuse-db
+sensitive_url=mozilla\.org
```
davehunt: Curious to find out why not `setup.cfg`? Do we have one for flake8? We should include that in the Travis setup too.

davehunt: Nit: Inconsistent use of whitespace surrounding `=`

davehunt:
Looks great, I added a few comments, and can't really provide much input for the app changes.

.travis.yml
bobsilverberg: Good catch. I need to set up a new bucket for these tests. Also, the secure variables in here will only work for my fork right now, so I should update those to work for the upstream repo.

.travis.yml
bobsilverberg: Done

.travis.yml
bobsilverberg: Heroku expects the requirements file to be in the root of the project and called `requirements.txt`. I could try using that base file to include other files, but I'm not sure if that's compatible with `peep`, plus I don't really want the test requirements installed on Heroku so I'd also need to do a conditional include, and I'm not sure if that's possible.

So while the bottom line is that I agree it would be nicer to do it as you suggest, I'm not going to allow that to block this effort. It is something I will look into in the future.

oneanddone/functional_tests/conftest.py
bobsilverberg: That _is_ better. I did not try removing the scope. I'll do that now.

bobsilverberg: I just tried without `scope='session'` and it generated an error:

`ScopeMismatch: You tried to access the 'function' scoped fixture 'base_url' with a 'session' scoped request object, involved factories`

oneanddone/functional_tests/requirements.txt
bobsilverberg: No. If we could I wouldn't have done this. ;)

I needed to hack `pytest-django` in order to make my tests work. Without this hack Django was not serving static assets via the live_server and therefore the functional tests that need JavaScript (which include any that need to log in), were not working.

I did raise an issue against `pytest-django` for this, and I should document that in the requirements file, so I will.

oneanddone/functional_tests/conftest.py
bobsilverberg: Good idea. Changed.

oneanddone/functional_tests/pages/base.py
bobsilverberg: Yes, these are identical to the tests in oneanddone-tests, which I will remove once this process is complete, so no need to review the tests themselves.

pytest.ini
bobsilverberg: I've moved the contents of `pytest.ini` into `setup.cfg`, and I agree that we should enable `flake8` tests on Travis. Running it locally I see there are a few issues, so I will make that a separate issue to enable those on Travis which can also address all the flake8 failures.

bobsilverberg:
Thanks for the review, @davehunt. I have addressed your comments in a new commit.

.travis.yml
davehunt: If you want to use requires.io then it looks for `req*.txt` in the root, and `requirements/*.txt`. So while a `requirements.txt` in the root is fine, I would still recommend a `requirements/test.txt` file. Unless you don't intend to monitor your dependencies.

oneanddone/functional_tests/conftest.py
davehunt: Your error is truncated. The involved factories are what I would need to know to help. :smile:

oneanddone/functional_tests/requirements.txt
davehunt: I'm surprised if One and Done is doing anything radically different form other apps using this plugin, but I don't know anything about the plugin or even Django itself. I'll follow the issue you've raised.

.travis.yml
bobsilverberg: Fair enough. That's a simple enough change for now, so I've moved the requirements into `requirements/test.txt`

oneanddone/functional_tests/conftest.py
bobsilverberg: Oops:

```
ScopeMismatch: You tried to access the 'function' scoped fixture 'base_url' with a 'session' scoped request object, involved factories
../../../.virtualenvs/oneanddone-functional-tests/lib/python2.7/site-packages/pytest_selenium/pytest_selenium.py:49:  def _verify_url(request, base_url)
oneanddone/functional_tests/conftest.py:30:  def base_url(base_url, live_server)
```

oneanddone/functional_tests/requirements.txt
bobsilverberg: I don't know why nobody else has run into this either. I opened the issue and if they fix it or suggest a resolution I will update the app accordingly.

oneanddone/functional_tests/conftest.py
davehunt: Right, the verify_url fixture is session scoped, and I was wrong about the default scope. I even have this explicitly set in pytest-selenium. Sorry for the noise.

davehunt:
Test related code looks good. I haven't run any tests locally (would you like me to?) This is an r+ from me, but would be good to get someone familiar with Django or pytest-django to look over the changes too.

bobsilverberg:
@akatsoulas or @glogiotatidis r? please and thank you

setup.cfg
```diff
@@ -1,4 +1,9 @@
 [flake8]
 ignore=E121,E123,E124,E125,E126,E129
 max-line-length=150
```
glogiotatidis: :exclamation:

glogiotatidis:
Great work @bobsilverberg! r+ :100:

stephendonner:
I like pushing green buttons which are backed by r+ code :-)

PR #311, opened 2015-12-16
Move all functional tests from oneanddone-tests into this repo

bobsilverberg:


bobsilverberg:
@davehunt r? on the changes to the `persona_test_user` fixture.

oneanddone/tests/functional/conftest.py
```diff
@@ -14,7 +16,18 @@ def capabilities(capabilities):
 
 @pytest.fixture
 def persona_test_user():
-    return requests.get('http://personatestuser.org/email').json()
+    max_retries = 5
+    attempt = 1
+    while attempt <= max_retries:
+        user_info = requests.get('http://personatestuser.org/email').json()
```
davehunt: This will raise a `ValueError` if the response is not valid JSON. We probably want to retry in that scenario.

bobsilverberg:
It's ready now @davehunt. Sorry for that.

oneanddone/tests/functional/conftest.py
```diff
@@ -14,7 +16,18 @@ def capabilities(capabilities):
...
+        if user_info.get('email'):
```
davehunt: Do we only care about the 'email' key? I guess that will likely catch the majority of issues.

davehunt:
I think we should catch ValueError for invalid JSON, but otherwise this looks good as a first pass. I'm not sure what the implications are of preventing pytest from capturing the log output though. Does it mean assertion failures will not be parsed?

bobsilverberg:
@davehunt I addressed your comment about `ValueError`. I also did a quick test by adding a failing `==` assertion and using `-s` does not seem to impede `pytest`'s ability to report assertion failure details. I am assuming the capturing of output is more of a convenience than a requirement.

davehunt:
r+ though by adding retries at the same time it might be a while before we see some information on what's failing. I think if we make this a tiny plugin then we should consider using the retry package as a dependency. Note that I haven't reviewed the tests as I assume they're identical to those already reviewed, and I also haven't run any tests. Let me know if you'd like to do either.

bobsilverberg:
Thanks @davehunt. I agree the `persona_test_user` fixture could be improved, and that a plugin would be a nice idea. I'm going to merge this so I can get the tests moved, but I've added creating that plugin to my TODO list.

PR #314, opened 2016-01-06
Allow for functional tests to be run against deployed environments

bobsilverberg:
Split out dependencies that are only needed for testing with a local host
Remove --reuse-db from setup.cfg
Make use of pytest-django fixtures conditional on tests running locally

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ env:
 install:
   - "pip install -r requirements.txt"
   - "pip install -r requirements/test.txt"
+  - "pip install -r requirements/test-local.txt"
   - "npm install"
   - "psql -c 'create database oneanddone;' -U postgres"
 
@@ -50,7 +51,7 @@ script:
   - "python -m smtpd -n -c DebuggingServer localhost:1025 &"
   - "flake8 ."
   - "python manage.py test oneanddone/tasks/ oneanddone/users/"
-  - "py.test -v -s --driver Firefox --html results/results.html oneanddone/tests/functional/"
+  - "py.test -v -s --driver Firefox --html results/results.html --reuse-db oneanddone/tests/functional/"
 
 notifications:
     irc:
--- a/oneanddone/tests/functional/conftest.py
+++ b/oneanddone/tests/functional/conftest.py
@@ -50,12 +50,24 @@ def new_user(persona_test_user):
 
 
 @pytest.fixture(scope='session')
-def base_url(base_url, live_server):
-    return base_url or live_server.url
+def base_url(base_url, request):
+    return base_url or request.getfuncargvalue("live_server").url
 
 
 @pytest.fixture(scope='function')
-def task(base_url, transactional_db):
-    if '127.0.0.1' in base_url or 'localhost' in base_url:
+def is_local(base_url):
+    return '127.0.0.1' in base_url or 'localhost' in base_url
+
+
+@pytest.fixture(scope='function', autouse=True)
+def use_transactional_db(base_url, is_local, request):
+    if is_local:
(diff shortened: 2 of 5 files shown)


bobsilverberg:
@davehunt I figured out how to invoke fixtures programatically, as we discussed, and it seems to work well. The tests are now passing locally, on Travis and on Jenkins. :)

You can see the most recent adhoc at http://webqa-ci-staging1.qa.scl3.mozilla.com:8080/job/oneanddone.adhoc/54/

The only odd thing is that the build is being marked as failed even though all the tests are passing. Is that because of the pytest-warnings that are being generated?

oneanddone/tests/functional/conftest.py
```diff
@@ -50,12 +50,24 @@ def new_user(persona_test_user):
 
 
 @pytest.fixture(scope='session')
-def base_url(base_url, live_server):
-    return base_url or live_server.url
+def base_url(base_url, request):
+    return base_url or request.getfuncargvalue("live_server").url
```
davehunt: Interesting, I was aware of this, but wasn't aware it invoked the fixture too! Awesome!

.travis.yml
```diff
@@ -36,6 +36,7 @@ env:
 
 install:
   - "pip install -r requirements.txt"
+  - "pip install -r requirements/test-local.txt"
```
davehunt: Is this still needed if we can invoke the fixtures programmatically? Does even just installing pytest-django cause problems?

davehunt:
> The only odd thing is that the build is being marked as failed even though all the tests are passing. Is that because of the pytest-warnings that are being generated?

This is likely the case, although I can't find any documentation that supports that warnings would cause a non-zero exit code. Your warnings are due to you using pytest 2.8 with outdated versions of pytest-xdist and pytest-timeout. Unfortunately you can't simply update both of these as the latest versions are not compatible with each other. For Bedrock, which is using pytest 2.8, I've removed the step to install infrastructure requirements, which you might want to do and then either add pytest-xdist into your requirements.txt or add a step to install it in the job.

In order to update our suites to pytest 2.8 I'm seriously considering dropping our use of pytest-timeout, and addressing the root cause of any builds that hang, though I know that might not be an easy task.

davehunt:
btw, r+ nice work!

.travis.yml
bobsilverberg: I think it does, but I'll give it another test now that I've made these changes.

bobsilverberg: Yep, the plugin tries to import `settings` from `django.conf` which fails without the app installed [1], so we need this.

[1] http://webqa-ci-staging1.qa.scl3.mozilla.com:8080/job/oneanddone.adhoc/55/console

bobsilverberg:
Thanks @davehunt. I removed the infrastructure requirements from the job and install `pytest-xdist` myself and that seems to have dealt with both the warnings and the reporting of the build as a failure. I will merge these changes soon and then set up our Jenkins prod to run tests from here.

requirements/test.txt
```diff
@@ -1,6 +1,4 @@
 bidpom
 pytest-selenium
+pytest-xdist
```
davehunt: You're not pinning pytest at all, which means future releases could break your tests. I think it might be a good idea to pin both pytest and pytest-xdist.

bobsilverberg: Good point. I'm pinning now.

bobsilverberg:
@davehunt Do you want to take a final look at this before I merge it?

davehunt:
nope, but you might want to squash before merging :)

bobsilverberg:
I will, thanks!

PR #323, opened 2016-01-28
Creating a PR for this react experiment for demo purposes

bobsilverberg:


oneanddone/tasks/static/js/tasklist.js
```diff
@@ -0,0 +1,415 @@
...
+                url={team.url}
+                key={team.id}
+                {...this.props} />
+        );
+      }
+    }, this);
+    return (
+      <div>
+        <span className="filter-heading">Teams</span>
+        <span className="filter-toggle" onClick={this.handleToggle}> {(this.state.expanded) ? '-':'+'} </span>
```
pdehaan: The `filter-heading` and `filter-toggle` sections are duplicated 3x. Not sure if you want to convert them into a new component, and maybe something like this:
```html
<ToggleHeading label="Projects" handleToggle={this.handleToggle} expanded={this.state.expanded} />

<span className="filter-heading">{this.props.label}</span>
<span className="filter-toggle" onClick={this.props.handleToggle}>{this.state.expanded ? '-' : '+'}</span>
```

You could possibly also expand it to include the surrounding `<div>` and `<ul>{nodes}</ul></div>` since those seem boilerplate as well.

oneanddone/tasks/static/js/tasklist.js
```diff
@@ -0,0 +1,415 @@
...
+      executionTime.push(e.target.value);
+    } else {
+      executionTime.splice($.inArray(e.target.value, executionTime), 1);
+    }
+    this.props.onFilterChange();
+  },
+  render: function() {
+    let etId = 'et' + this.props.executionTime;
+    return (
+      <label htmlFor={etId}>
```
pdehaan: You may want to add a bit of padding here. I was hacking the dev tools and adding a 10-15px of `padding-right` seemed to space these out a bit.

oneanddone/tasks/static/js/tasklist.js
```diff
@@ -0,0 +1,415 @@
...
+})
+
+
+ReactDOM.render(
+    <TaskPage url="/api/v1/task"
+              team_url="/api/v1/taskteam"
+              project_url="/api/v1/taskproject"
+              type_url="/api/v1/tasktype" />,
+    document.getElementById('content')
+);
```
pdehaan: Not sure if this would get any easier to read if you moved from one major React component to a few smaller ones:

```js
ReactDOM.render(
  <div className="taskPage">
    <div className="task-filters">
      <form method="get" className="filters">
        <fieldset>
          <SearchFilter />
          <ETFilter executionTimes={[15, 30, 45, 60]} />
          <TeamFilter type_url="/api/v1/taskteam" />
          <ProjectFilter type_url="/api/v1/taskproject" />
          <TypeFilter type_url="/api/v1/tasktype" />
        </fieldset>
      </form>
    </div>

    <TaskList data={this.state.data} onLoadMore={this.loadMoreTasks} moreTasks={this.state.moreTasks} />
  </div>,

  document.getElementById('content')
);
```

Not sure if that'd work, or require massive refactoring, but it'd move props closer to their intended targets so you wouldn't need to use a massive `componentDidMount()` to load all your AJAX data.
Now each filter could use it's own `componentDidMount` cycle to load its data and hopefully remove a few of the `{...this.props}` proxies. It may also require a bit less global `this.setState()` calls and move state management into the individual components instead of the global component.

oneanddone/tasks/static/js/tasklist.js
bobsilverberg: Thanks Peter. I did try to so a refactor similar to what you suggest, but I ended up getting errors about `this.state` being undefined. There's probably a way to do it. Maybe we can try hacking on it together at some point.

oneanddone/tasks/static/js/tasklist.js
bobsilverberg: Good idea. I'll give that a try. I haven't really been devoting any effort to making any of this pretty yet. I figured I'd take that on once it's all working exactly as I'd like.

oneanddone/tasks/static/js/tasklist.js
bobsilverberg: Interesting idea, and definitely worth a try. Thanks @pdehaan!

rbillings:
@bobsilverberg - is there anything else to be done with this, as you've already finished the demo?

bobsilverberg:
No, this can be closed, @rbillings.

edmorley:
Is the review app (https://dashboard.heroku.com/apps/oneanddone-dev-pr-323) still required, or can it be deleted? :-)

PR #325, opened 2016-02-01
Un-pin Selenium to figure out what's making Travis angry

bobsilverberg:


bobsilverberg:
@davehunt Here's where I'm going to try to figure out what's going on on Travis. It wouldn't upload the report for the PR, but it did for the commit, so that works.

FTR, the report is at https://s3.amazonaws.com/oneanddone-tests/mozilla/oneanddone/716/716.1/results.html and I'm going to take a look at it to see what might be going on.

bobsilverberg:
@davehunt Two of the failures seem to be the "failure to click" issue. The screenshots show that the link was meant to be clicked (it's highlighted) but that the click never happened (we're still on that screen). 

The other failure is that new exception that happens when you try to click the wrong element. I'm not sure how to get around it though, as I believe I am targeting the proper element, but I guess the browser/selenium feels that the surrounding div is on top of that element. Here's the markup:

```html
<div class="actions-container">
     <a href="/profile/delete/" id="delete-profile" class="button">Delete Profile</a>
      <button type="submit" id="save-profile" class="button">Save</button>
</div>
```

and my locator is: 

`_delete_profile_button_locator = (By.ID, 'delete-profile')` 

which seems correct to me.

I'm going to see if I can request a more recent version of Firefox on Travis and whether that addresses the click issues, but I'm not sure what to do about the wrong element issue.

bobsilverberg:
For completeness, here's the exception from WebDriver when trying to click the delete button:

`WebDriverException: Message: Element is not clickable at point (501.6000061035156, 68.78334045410156). Other element would receive the click: <div class="actions-container"></div>`

bobsilverberg:
@davehunt Even with Firefox 44 I am having no luck. I am getting both the failure to click problem as well as the wrong element problem. The latest report is at https://s3.amazonaws.com/oneanddone-tests/mozilla/oneanddone/720/720.1/results.html. Any suggestions you have would be appreciated!

bobsilverberg:
I tried monkeying around with z-indexes to see if I could change the order of the elements, but that didn't help. Also, I'm not seeing any of these failures locally with Fx 44 and Se 2.50.1, so I'm not really sure what's going on!

davehunt:
Perhaps it's a timing/rendering issue. Does adding a sleep help? It's at least good that we have a reasonably reproducible test case, even if it is on Travis! I believe there was a change in Selenium to attempt to click the center of an element, and if that fails to attempt to click each corner, but I did see a report that clicking the corners of a rounded element caused other problems.

davehunt:
Is it also possible that the element needs to be scrolled into view? It could be that the window is too small by default on Travis? Perhaps dump the window size to console?

bobsilverberg:
@davehunt Changing the default screen size for Xvfb fixed the whole problem!

davehunt:
Interesting, thanks @bobsilverberg! I wonder if we need to specify a resolution or Window size for our Jenkins jobs, to ensure they're consistent. Let's keep this is mind for when we see similar failures.

PR #327, opened 2016-02-03
Bug 1244123 - Convert functional tests to use PyPOM

bobsilverberg:


bobsilverberg:
Also replaced `type_x` methods with setters.

bobsilverberg:
@davehunt r?

oneanddone/tests/functional/pages/tasks/whats_next.py
```diff
@@ -6,6 +6,4 @@
 
 class WhatsNextPage(Base):
```
davehunt: Is this page object even needed now?

oneanddone/tests/functional/pages/user/user_profile_edit.py
```diff
@@ -22,45 +20,49 @@ class UserProfileEditPage(Base):
 
     @property
     def bugzilla_email(self):
-        return self.selenium.find_element(
-            *self._bugzilla_email_input_locator).get_attribute('value')
+        return self.find_element(
+            self._bugzilla_email_input_locator).get_attribute('value')
+
+    @bugzilla_email.setter
```
davehunt: After PyPOM, maybe this is the next improvement to make across our suites. :smile:

davehunt:
r+ I like it!

oneanddone/tests/functional/pages/tasks/whats_next.py
bobsilverberg: We do end up at that page in one of the tests, but I guess there are no assertions about it, otherwise it would have properties. I think before we were just asserting that we were on the correct page, which was done via page titles, but I removed those. I realize now that that may result in lost coverage. I'll take a look.

bobsilverberg: We do end up at that page in one of the tests, but I guess there are no assertions about it, otherwise it would have properties. I think before we were just asserting that we were on the correct page, which was done via page titles, but I removed those. I realize now that that may result in lost coverage. I'll take a look.

PR #332, opened 2016-03-03
Bug 1252540 - Django security release 2016-03-01 (One and Done)

karlht:
Here you go, Bob.  Thanks very much.


bobsilverberg:
Looks good, thanks @karlht!  Let's wait for Travis to be happy then we can merge. r+

PR #333, opened 2016-03-13
Removed section that redirects for functional tests

rbillings:


stephendonner:
r+ - thanks!

bobsilverberg:
@rbillings I would suggest that rather than deleting this section, just change it to point at the readme which is currently located in https://github.com/mozilla/oneanddone/tree/master/oneanddone/tests/functional

PR #335, opened 2016-03-20

VarnaSuresh:
I have hard-coded the admin user into the function written to create a task. Unsure if that's the best way to go about it. Please review my work.


VarnaSuresh:
The Travis CI build is failing because the admin user I have hard-coded into the test doesn't have admin privileges on the testing site.

oneanddone/tests/functional/tests/test_task_assignment.py
```diff
@@ -0,0 +1,57 @@
...
+class TestAvailableTasks():
+    @pytest.mark.nondestructive
+    def test_create_tasks(self, base_url, selenium, task, random_title):
+
+        admin_user = {'email': "varna.suresh@outlook.com",
+                      'password': "varna123",
+                      'name': "Varna",
+                      'username': "Varna",
+                      'url': 'http://www.mozilla.org/'
+        }
```
davehunt: Rather than hard-coding an admin user, please try creating an admin user for the purposes of the test. The documentation here http://pytest-django.readthedocs.org/en/latest/helpers.html#admin-user-a-admin-user-superuser may help. Also, please change your Persona account password as soon as possible, so that it is not exploited by the credentials being shared in plain text here.

oneanddone/tests/functional/pages/home.py
```diff
@@ -51,3 +63,27 @@ def click_pick_a_task_button(self):
     def click_available_tasks(self):
         self.find_element(self._available_tasks_locator).click()
         return AvailableTasksPage(self.selenium, self.base_url).wait_for_page_to_load()
+
+    def click_create_task(self, task_name):
+        self.find_element(self._create_task_button).click()
+        assert self.is_element_displayed(self._find_task_title)
```
davehunt: asserts should be in the tests, not the page objects

oneanddone/tests/functional/pages/home.py
```diff
@@ -51,3 +63,27 @@ def click_pick_a_task_button(self):
...
+        self.find_element(self._find_task_team).click()
+        assert self.is_element_displayed(self._choose_team)
+        self.find_element(self._choose_team).click()
+        assert self.is_element_displayed(self._instruction_locator)
+        self.find_element(self._instruction_locator).click()
+        instructions = self.find_element(self._find_task_instructions)
+        instructions.send_keys("This is a non repeatable task")
+        assert self.is_element_displayed(self._find_task_repeatable)
+        self.find_element(self._find_task_repeatable).click()
+        time.sleep(5)
```
davehunt: we should avoid sleeps, and instead use explicit waits using `WebDriverWait`

oneanddone/tests/functional/pages/home.py
```diff
@@ -51,3 +63,27 @@ def click_pick_a_task_button(self):
     def click_available_tasks(self):
         self.find_element(self._available_tasks_locator).click()
         return AvailableTasksPage(self.selenium, self.base_url).wait_for_page_to_load()
+
+    def click_create_task(self, task_name):
```
davehunt: This is doing more than just clicking create task. It's not a problem to have a function that combines several interactions, but we should make sure the name is clear. In this case I think we're doing too much in the page object, and if we break these interactions down we could call them directly from the test.

oneanddone/tests/functional/tests/test_task_assignment.py
```diff
@@ -0,0 +1,57 @@
...
+
+        home_page = HomePage(selenium, base_url).open()
+        assert not home_page.is_user_logged_in
+        edit_profile = home_page.login(admin_user)
+        assert edit_profile.is_user_logged_in
+        home_page.click_available_tasks()
+        home_page.click_create_task(random_title)
+
+    @pytest.mark.nondestructive
+    def test_assign_tasks(self, base_url, selenium, task, new_user, random_title):
```
davehunt: This test depends on `test_create_tasks` succeeding. We can't guarantee this test will even run first. It's good practice for tests to take care of all data they rely on, and can run in isolation. This should be a single test that logs in, creates a task, assigns it to one user, and then checks that another user cannot take it. The creation and assignment of the task can be taken care of in a similar way to https://github.com/mozilla/oneanddone/blob/master/oneanddone/tests/functional/conftest.py#L69-L73

oneanddone/tests/functional/tests/test_task_assignment.py
davehunt: If you're unable to log in or access certain areas of the site it may be due to some changes we've made due to these credentials being shared. Let me or someone in #mozwebqa on IRC know if you're seeing issues and we can work to resolve them.

rbillings:
@davehunt this is ready for review when you have time

oneanddone/tests/functional/conftest.py
```diff
@@ -59,6 +59,11 @@ def is_local(base_url):
     return '127.0.0.1' in base_url or 'localhost' in base_url
 
 
+@pytest.fixture(scope='session')
+def is_local2(base_url):
+    return '127.0.0.1' in base_url or 'localhost' in base_url
```
davehunt: You should be able to make the existing `is_local` session scoped, as the value will not change between tests. That way you won't need to duplicate this code.

oneanddone/tests/functional/conftest.py
```diff
@@ -71,3 +76,10 @@ def task(base_url, is_local):
     if is_local:
         from oneanddone.tasks.tests import TaskFactory
         return TaskFactory.create()
+
+
+@pytest.fixture(scope='session')
```
davehunt: Why have you chosen session scope? This means that multiple tests using this fixture in a single session will use the same task. It would be better for all tests to create their own task.

oneanddone/tests/functional/conftest.py
```diff
@@ -71,3 +76,10 @@ def task(base_url, is_local):
...
+def nonrepeatable_task(base_url, is_local2):
```
davehunt: This fixture does not use `base_url` so we can remove it. Also, can we rename it to `non_repeatable_task`?

oneanddone/tests/functional/pages/home.py
```diff
@@ -18,6 +19,7 @@ class HomePage(Base):
     _suggested_first_tasks_heading_locator = (By.CSS_SELECTOR, '.task-list-container h3')
     _suggested_first_task_locator = (By.CSS_SELECTOR, '.task-list > li')
     _available_tasks_locator = (By.ID, 'available-tasks')
+    _find_search_button = (By.CSS_SELECTOR, '#id_search')
```
davehunt: Add a `_locator` suffix. Also, this could be a `By.ID`, but there's no need to update this.

oneanddone/tests/functional/tests/test_task_assignment.py
```diff
@@ -0,0 +1,41 @@
...
+import pytest
+
+from pages.home import HomePage
+
+
+class TestAvailableTasks():
+    @pytest.mark.nondestructive
+    def test_assign_tasks(self, base_url, selenium, nonrepeatable_task, new_user):
+        home_page = HomePage(selenium, base_url).open()
+        assert not home_page.is_user_logged_in
```
davehunt: There's no need to check that we're not logged in. This is covered by another test.

oneanddone/tests/functional/tests/test_task_assignment.py
```diff
@@ -0,0 +1,41 @@
...
+
+        edit_profile = home_page.login(new_user)
+        assert edit_profile.is_user_logged_in
```
davehunt: As above, no need to check this here.

oneanddone/tests/functional/tests/test_task_assignment.py
```diff
@@ -0,0 +1,41 @@
...
+
+        available_tasks_page = home_page.click_available_tasks()
+        assert available_tasks_page.is_available_tasks_list_visible
+        assert len(available_tasks_page.available_tasks) > 0
```
davehunt: These checks are not needed, as they're covered elsewhere.

oneanddone/tests/functional/tests/test_task_assignment.py
```diff
@@ -0,0 +1,41 @@
...
+
+        nonrepeatable_task = available_tasks_page.available_tasks[0]
+        task_name = nonrepeatable_task.name
+        task_details = nonrepeatable_task.click()
+        assert task_name == task_details.name
+        assert task_details.is_get_started_button_visible
+        task_details.click_get_started_button()
+        assert task_details.is_abandon_task_button_visible
+        assert task_details.is_complete_task_button_visible
+        assert task_details.is_save_for_later_button_visible
```
davehunt: This assigns the task to the logged in user, but as far as I can tell it doesn't verify that the task is non-repeatable.

oneanddone/tests/functional/tests/test_task_assignment.py
```diff
@@ -0,0 +1,41 @@
...
+
+    @pytest.mark.nondestructive
+    def test_reassign_task(self, base_url, selenium, nonrepeatable_task, new_user):
+        home_page = HomePage(selenium, base_url).open()
+        assert not home_page.is_user_logged_in
+        edit_profile = home_page.login(new_user)
+        assert edit_profile.is_user_logged_in
+        available_tasks_page = home_page.click_available_tasks()
+        home_page.search_for_task(nonrepeatable_task.name)
+        assert len(available_tasks_page.available_tasks) == 0
```
davehunt: This checks that the non-repeatable task created in the fixture cannot be found via search, but we don't have a positive test for this.

oneanddone/tests/functional/conftest.py
davehunt: I see. We shouldn't make tests dependent on data they don't set up themselves though. Each test should be able to run individually, and tests should be able to execute in any order. If a test requires a tasks to already be assigned, then it it either needs to assign it via the UI, or include this as part of the task creation.

oneanddone/tests/functional/tests/test_task_assignment.py
VarnaSuresh: Only tasks that can be assigned to that particular user will appear as results for the search. Because that task has already been assigned in the previous test, I am checking if the task is listed.

oneanddone/tests/functional/tests/test_task_assignment.py
VarnaSuresh: I will add that. As I created the task as a non repeatable one, I assumed it did not have to be tested.

davehunt: Other tests are creating tasks - how can you be sure that the first task is the one created in your fixture?

oneanddone/tests/functional/tests/test_task_assignment.py
davehunt: I think that's okay, but we need a positive check to be sure that this is doing what we're expecting. What I mean is that we need a test that searches for an expected task and finds it.

rbillings:
@VarnaSuresh - are you still working on this?

VarnaSuresh:
@rbillings Yes I am

oneanddone/tests/functional/conftest.py
VarnaSuresh: To be able to both assign the task to a user and have a new user check if it can be assigned to him  in the same test, a logout and new login should happen in the same test right? How can I do that? Alternatively, 'assignee' as such doesn't seem to be an attribute of a task. So how can I assign the task at creation without using the UI?

davehunt: I'm not familiar enough with Django to help with creating a task that has an assignee. You could do it via the UI but that would add to the test duration and perhaps make it less robust. Looking at `TaskFactory` it seems the owner is set.. perhaps there's a way to set an assignee? https://github.com/mozilla/oneanddone/blob/master/oneanddone/tasks/tests/__init__.py#L65-L81

davehunt: On further inspection we may need to create a task that is non-repeatable and has a started task attempt.

davehunt: This looks like it might be an example of creating such a task: https://github.com/mozilla/oneanddone/blob/7195406ba41fa8e2178b06a9ee1c475f3322749e/oneanddone/tasks/tests/test_api.py#L64-L97

rbillings:
@VarnaSuresh - do you have an ETA on the next update?

VarnaSuresh:
@rbillings Hey, I have closed this request and opened a new one with my latest changes.

rbillings:
Great, thanks @VarnaSuresh

PR #345, opened 2016-05-23
Updated to support latest changes in PyPOM

davehunt:
@rbillings r?


rbillings:
Changes look good, R+

PR #348, opened 2016-06-06
pin PyBrowserID

m8ttyB:
PyBrowerID was recently updated which broke our builds.

```
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django_browserid/base.py", line 252, in <module>

    from browserid.errors import Error as PyBrowserIDError

ImportError: No module named errors
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -128,6 +128,9 @@ psycopg2==2.6.1
 # sha256: rxL5lFTolenkMMdXI8O5QZwLzNDqd4A2YtaprFb2Qls
 py-bcrypt==0.3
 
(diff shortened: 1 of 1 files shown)


rbillings:
I was able to install the updated requirements without error, R+

davehunt:
r+ looks good. I believe it's good practice to list all dependencies so that you don't have a downstream dependency update that affects your application (perhaps there are more?). You can add `--no-deps` to ensure that you only install the explicitly listed dependencies (which should all be pinned). See https://pip.pypa.io/en/stable/user_guide/#ensuring-repeatability for more.

m8ttyB:
Thanks for the R. Merging

PR #353, opened 2016-07-06

VarnaSuresh:


VarnaSuresh:
@davehunt Hey, could you review my latest changes. Thank you.

oneanddone/tests/functional/tests/test_task_assignment.py
```diff
@@ -0,0 +1,22 @@
+# This Source Code Form is subjectfrom django import forms to the terms of the Mozilla Public
```
davehunt: This license block is incorrect - it has an import inside it.

oneanddone/tests/functional/tests/test_task_assignment.py
```diff
@@ -0,0 +1,22 @@
...
+        home_page = HomePage(selenium, base_url).open()
+        home_page.login(new_user)
+        available_tasks_page = home_page.click_available_tasks()
+
+        # Check if assignable task is found
+        home_page.search_for_task(task.name)
+        assert len(available_tasks_page.available_tasks)
+
+        home_page.search_for_task(nonrepeatable_assigned_task.name)
+        assert len(available_tasks_page.available_tasks) == 0
```
davehunt: There are really two separate tests here. I would call one something like `test_that_unassigned_task_is_available` and the other `test_that_assigned_task_is_not_available`. Could you also add to the assigned task test a login as the assigned user and an assert that the task is listed on their dashboard? Without this, the test could be a false positive as the test may not exist at all.

davehunt:
This is looking good @VarnaSuresh, just a couple of comments. Sorry it took so long to review it.

VarnaSuresh:
@davehunt Hey I have made all those changes. When I updated by branch the Travis build failed though

VarnaSuresh:
@davehunt Please review my changes. Everything seems to be working ok.

oneanddone/tests/functional/tests/test_task_assignment.py
```diff
@@ -0,0 +1,27 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from pages.home import HomePage
+
+
+class TestAvailableTasks():
+
+    @pytest.mark.nondestructive
```
davehunt: These tests are both destructive as they create tasks.

oneanddone/tests/functional/tests/test_task_assignment.py
```diff
@@ -0,0 +1,27 @@
...
+    def test_assigned_task_is_not_available(self, base_url, selenium, assigned_task, new_user):
+        home_page = HomePage(selenium, base_url).open()
+        home_page.login(new_user)
+        available_tasks_page = home_page.click_available_tasks()
+        home_page.search_for_task(assigned_task['task'].name)
+        assert len(available_tasks_page.available_tasks) == 0
+        assert assigned_task['task'].is_available_to_user(assigned_task['user']) == True
+
+    @pytest.mark.nondestructive
```
davehunt: As above, this is a destructive test.

oneanddone/tests/functional/tests/test_task_assignment.py
```diff
@@ -0,0 +1,27 @@
...
+class TestAvailableTasks():
+
+    @pytest.mark.nondestructive
+    def test_assigned_task_is_not_available(self, base_url, selenium, assigned_task, new_user):
+        home_page = HomePage(selenium, base_url).open()
+        home_page.login(new_user)
+        available_tasks_page = home_page.click_available_tasks()
+        home_page.search_for_task(assigned_task['task'].name)
+        assert len(available_tasks_page.available_tasks) == 0
+        assert assigned_task['task'].is_available_to_user(assigned_task['user']) == True
```
davehunt: I don't think we need this assert.

oneanddone/tests/functional/tests/test_task_assignment.py
```diff
@@ -0,0 +1,27 @@
...
+
+    @pytest.mark.nondestructive
+    def test_that_unassigned_task_is_available(self, base_url, selenium, task, new_user):
+        home_page = HomePage(selenium, base_url).open()
+        home_page.login(new_user)
+        available_tasks_page = home_page.click_available_tasks()
+        home_page.search_for_task(task.name)
+        assert len(available_tasks_page.available_tasks)
```
davehunt: Can we assert that there is exactly one task with this name?

oneanddone/tests/functional/tests/test_task_assignment.py
VarnaSuresh: @davehunt  The second assert checks whether the task is available to the user that created it and the first one checks that its unavailable to the new user. Don't we require both checks?

davehunt: The second assert is not checking anything on the front-end or anything that is a consequence of interacting with the front-end, it's more of a unit test and should be covered elsewhere accordingly. If you want to check that the task is available via the front-end you would need to login as the assigned user and verify that you can find the task.

oneanddone/tests/functional/tests/test_task_assignment.py
```diff
@@ -22,4 +22,5 @@ def test_that_unassigned_task_is_available(self, base_url, selenium, task, new_u
         home_page.login(new_user)
         available_tasks_page = home_page.click_available_tasks()
         home_page.search_for_task(task.name)
-        assert len(available_tasks_page.available_tasks)
+        available_tasks = x.name for x in available_tasks_page.available_tasks
+        assert available_tasks.count(task.name) == 1
```
davehunt: Aren't the task names unique? Searching for the unique task name should allow us to verify the length of available tasks is 1. This is failing in Travis CI - did you run it locally before pushing?

oneanddone/tests/functional/tests/test_task_assignment.py
```diff
@@ -0,0 +1,24 @@
...
+        assert len(available_tasks_page.available_tasks) == 0
+        assert assigned_task['task'].is_available_to_user(assigned_task['user']) == True
+
+    def test_that_unassigned_task_is_available(self, base_url, selenium, task, new_user):
+        home_page = HomePage(selenium, base_url).open()
+        home_page.login(new_user)
+        available_tasks_page = home_page.click_available_tasks()
+        home_page.search_for_task(task.name)
+        available_tasks = [x.name for x in available_tasks_page.available_tasks]
+        assert available_tasks.count(task.name) == 1
```
davehunt: Why not simply `assert len(available_tasks_page.available_tasks) == 1`

VarnaSuresh: Because the search page has all tasks that has the task.name as a substring. Which might be more than one.

davehunt: If that's the case, doesn't it mean that the assert in `test_assigned_task_is_not_available` could fail, as there may be other tasks that match that substring?

oneanddone/tests/functional/tests/test_task_assignment.py
VarnaSuresh: Okay. So maybe I can create a third test that performs this login and checks if its on his dashboard?

oneanddone/tests/functional/tests/test_task_assignment.py
VarnaSuresh: Yeah. I will change that as well.

oneanddone/tests/functional/tests/test_task_assignment.py
davehunt: You can do this in a follow-up pull request if you prefer.

oneanddone/tests/functional/tests/test_task_assignment.py
davehunt: Isn't your code the same though? It's counting the number of times the task name appears in the list of available tasks. Under what circumstances would a task name exist as a substring from another task? Perhaps we should make the tasks names udids or similar?

davehunt:
Did you mean to close this @VarnaSuresh?

VarnaSuresh:
@davehunt Hey. Yeah I will open a new pull request once I am done with the changes.

VarnaSuresh:
@davehunt Hey. Is there a way to login through the UI as a user created using UserFactory without explicitly setting up the password details in conftest.py?

rbillings:
@VarnaSuresh you can learn more about logging in & Persona testing here: https://github.com/mozilla/bidpom/wiki
There is no need to save the login variables.

PR #357, opened 2016-09-06
Update pytest and request

stephendonner:
Specifically, grab the latest for requests, but version-match for pytest (currently at 3.0.2, and will be picked up with this change), so we automatically pick up any point-release versions.

@rbillings / @karlht r?


karlht:
Looks fine to me.  r+

stephendonner:
Thx; merging in.

PR #358, opened 2016-09-07

VarnaSuresh:
I changed most of the test, so I have opened a new pull request. 

There are two tests I have written. One that checks that an assigned, one time task is not available and one that checks if an unassigned one time task is available to a user. The second test goes on to assign the task as well, and check if it is available to the user who has taken it.

@davehunt Please review it


davehunt:
Thank you for your contribution, @VarnaSuresh, I'm really impressed with your work! I understand that there's a possibility the One and Done site might be shut down in the future, so I don't think there's going to be any more tests to work on here. If you're interested, I'd love find something else you could work on.

VarnaSuresh:
Hey @davehunt! Contributing to this project was a fun learning experience for me. Rebecca has sent me a link to a bunch of potential projects. I shall take a look at it soon and get to work. Thank you for all your help!

PR #359, opened 2016-09-09
Skip tests that require a local instance to run

davehunt:


davehunt:
Merging this because it's just a couple of lines and should restore our Jenkins jobs.

